PR #43, opened 2026-06-30
fix(compress): raise maxSummaryLengthHard default to 3000, fix reject feedback

### LINKED ISSUE #42: [Bug]: compress always failed
### Bug Description

why on models always looping when i tried to compress previously it just work, i'm using v1.6.1(latest)
Image

### Expected Behavior

it should work like previous compress done not red

### Debug Context Logs

```json
there is no logs i can find you can just check on screenshoot i provide before
```

### Tool Call Details

_No response_

### DCP Version

1.6.1

### Opencode Version

1.17.11

### Model

Other (specify in description)

### Additional Context

GLM 5.2

ranxianglei:
## Summary
Fixes #42 (prihuda's report: `maxSummaryLengthHard=800` caused legitimate compressions to always fail).

- **`maxSummaryLengthHard` default 800 → 3000.** The 800 ceiling was miscalibrated — real summaries routinely run 800–3000 chars, so the safety net was catching normal compression instead of pathological bloat. The soft target (`maxSummaryLength`) stays at 200 to keep summaries concise (now that `search_context` can retrieve compressed detail, over-long summaries are wasteful).
- **Fix the reject feedback.** On failure the message used to say "aim for <=200" (the soft target) — misleading at failure time. It now guides the model to trim under the actual ceiling and keep only essential detail (conclusions, file paths, decisions, exact values).

## Verification
- `tsc --noEmit`: clean
- Docker `node:20` full suite: 496 pass / 0 fail

## Notes
Tuning will be revisited once the compression-bundle feature lands; for now this unblocks users hitting the default.

### DIFF
--- a/lib/compress/message.ts
+++ b/lib/compress/message.ts
@@ -55,7 +55,7 @@ export function createCompressMessageTool(ctx: ToolContext): ReturnType<typeof t
             for (const entry of input.content) {
                 if (entry.summary.length > maxSummaryLengthHard) {
                     throw new Error(
-                        `Summary too long (${entry.summary.length} chars, hard ceiling ${maxSummaryLengthHard}). Aim for <=${ctx.config.compress.maxSummaryLength}; exceed only when strictly necessary. Rewrite more concisely.`,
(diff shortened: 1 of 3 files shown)


ranxianglei:
link https://github.com/ranxianglei/opencode-acp/issues/42

PR #94, opened 2026-07-10
feat: gate notification wrap text behind debug flag

### LINKED ISSUE #84: [Bug]: Compression notification injected into chat dialog pollutes model context (1.10.0)
### Bug Description

**Describe the bug**

After upgrading to `opencode-acp@1.10.0`, the compression notification (e.g. `▣ ACP | Context 114.1K →112.2K — Compression #6`) is injected into the chat dialog as a message wrapped in `[ACP system message — not a user comment]` tags, and is **visible to the model** in its conversation context.

This pollutes the AI's context window with internal control-flow metadata on every compression cycle, wasting tokens and occasionally confusing the model into treating the bracketed text as a user instruction.

Observed in the chat dialog (screenshot below):

```
[ACP system message — not a user comment]

▣ ACP | Context 114.1K →112.2K — Compression #6 →b6

[ACP system message — not a user comment]
```

The notification appears as a full message bubble in the dialog, sandwiched between real conversation turns, and the model reads it as part of its input.

### Root cause

In `lib/ui/notification.ts`, `sendCompressNotification()` routes through `sendIgnoredMessage()` (compiled: `dist/index.js:4656`) when `config.pruneNotificationType === "chat"` (the default).

`sendIgnoredMessage()` calls `client.session.prompt()` with `parts: [{ type: "text", text: wrappedText, ignored: true }]`. The `ignored: true` flag is meant to signal "do not reply", but:

1. The message **still enters the session message stream** and is stored as a turn in the conversation.
2. The wrapped text (`[ACP system message — not a user comment]\n\n{notification}\n\n[ACP system message — not a user comment]`) is **included in the model's context window** on the next inference — the `ignored` flag only suppresses an immediate reply, it does not strip the text from the model's input.

So every compression cycle injects ~60-120 tokens of internal metadata into the model's visible context, and these accumulate across a long session.

The `toast` path (`config.pruneNotificationType === "toast"`) correctly uses `client.tui.showToast()` and never touches the conversation surface — but the **default is `"chat"`**, so most users hit the polluting path.

### Reproduction

1. Install `opencode-acp@1.10.0` with default config (`pruneNotificationType: "chat"`)
2. Start a long session (100K+ tokens) that triggers at least one compression
3. After compression completes, observe the chat dialog
4. A message bubble containing `[ACP system message — not a user comment]` + the `▣ ACP | Context ... →... — Compression #N` line appears in the dialog
5. On the next model turn, the model's context includes this text — it is not stripped despite the `ignored: true` flag

### Expected Behavior

Compression notifications should **never** appear in the model's conversation context. They are UI feedback for the human, not context for the model.

Two acceptable fixes:

1. **Change the default** `pruneNotificationType` from `"chat"` to `"toast"` — notifications go to the TUI toast, never the conversation.
2. **Or** make the `chat` path truly invisible: route through a mechanism that does not persist the text into the session message store at all (not just `ignored: true`, which only suppresses the reply but still stores + sends the text).

Option 1 is the smaller change and matches what most users actually want.

### Environment

- OpenCode version: 1.17.13
- opencode-acp plugin version: 1.10.0 (npm `opencode-acp@latest`)
- OS: Windows 10 (WSL2)
- `pruneNotificationType`: `"chat"` (default)

### Proposed Fix

**Minimal fix** — flip the default:

```diff
// lib/config.ts (defaultConfig)
- pruneNotificationType: "chat",
+ pruneNotificationType: "toast",
```

**Deeper fix** — if the `chat` path is kept for users who want inline notifications, the `sendIgnoredMessage` path should use a message-part type that is excluded from the model's context construction (e.g. a metadata-only part), not a `text` part with `ignored: true`. The current `ignored` flag does not prevent the text from reaching the model — it only prevents an auto-reply.

### Additional Context

Screenshot evidence (attached) shows the bracketed system message appearing as a full dialog bubble between real turns. Over a 6-compression session, this adds ~600-800 tokens of pure noise to the model's context — partially defeating the purpose of the compression that triggered it.

Related to the class of "plugin control-flow leaking into conversation surface" bugs (cf. the `console.warn` leak in #67, the `sentinel throw` / `output.parts` leak fixed in v1.4.x). The principle is the same: **anything the model can read in its context window should be real conversation content, not plugin-internal UI feedback.**

mangxy:
## Problem

Notifications (compression results, /acp command outputs) are wrapped with `[ACP system message — not a user comment]` text on both sides. This adds visual noise for end users in normal (non-debug) usage. Issue: #84

## Solution

Gate the wrapper text behind `config.debug`:
- `debug: false` (default) → clean notification text, no wrapper
- `debug: true` → wrapper added (useful during development to distinguish system vs user messages)

## Changes

```
lib/ui/notification.ts       | sendIgnoredMessage: +config param, conditional wrap
lib/commands/context.ts      | +PluginConfig import, +config field, pass config
lib/commands/decompress.ts   | +PluginConfig import, +config field, pass config ×7
lib/commands/help.ts         | pass config
lib/commands/manual.ts       | pass config
lib/commands/recompress.ts   | +PluginConfig import, +config field, pass config ×7
lib/commands/stats.ts        | +PluginConfig import, +config field, pass config
lib/commands/sweep.ts        | pass config ×3
tests/compression-groups.test.ts | +config in 5 handleDecompress/Recompress calls
```

## Verification

- `tsc --noEmit` ✅ no errors
- `npm test` ✅ 608/608 pass

Closes #84

### DIFF
--- a/lib/commands/context.ts
+++ b/lib/commands/context.ts
@@ -42,6 +42,7 @@
 
 import type { Logger } from "../logger"
 import type { SessionState, WithParts } from "../state"
+import type { PluginConfig } from "../config"
 import { sendIgnoredMessage } from "../ui/notification"
 import { formatTokenCount } from "../ui/utils"
 import { isIgnoredUserMessage } from "../messages/query"
@@ -55,6 +56,7 @@ export interface ContextCommandContext {
     logger: Logger
     sessionId: string
     messages: WithParts[]
+    config: PluginConfig
 }
 
 interface TokenBreakdown {
@@ -294,12 +296,12 @@ function formatContextMessage(breakdown: TokenBreakdown): string {
 }
 
 export async function handleContextCommand(ctx: ContextCommandContext): Promise<void> {
-    const { client, state, logger, sessionId, messages } = ctx
+    const { client, state, logger, sessionId, messages, config } = ctx
 
     const breakdown = analyzeTokens(state, messages)
 
     const message = formatContextMessage(breakdown)
 
     const params = getCurrentParams(state, messages, logger)
-    await sendIgnoredMessage(client, sessionId, message, params, logger)
+    await sendIgnoredMessage(client, sessionId, message, params, logger, config)
 }
--- a/lib/commands/decompress.ts
+++ b/lib/commands/decompress.ts
@@ -1,5 +1,6 @@
 import type { Logger } from "../logger"
 import type { SessionState, WithParts } from "../state"
+import type { PluginConfig } from "../config"
 import { syncCompressionBlocks } from "../messages"
 import { getCurrentParams } from "../token-utils"
 import { sendIgnoredMessage } from "../ui/notification"
@@ -22,6 +23,7 @@ import { saveSessionState } from "../state/persistence"
 export interface DecompressCommandContext {
     client: any
     state: SessionState
+    config: PluginConfig
     logger: Logger
     sessionId: string
     messages: WithParts[]
@@ -86,7 +88,7 @@ function formatAvailableBlocksMessage(availableTargets: CompressionTarget[]): st
 }
 
 export async function handleDecompressCommand(ctx: DecompressCommandContext): Promise<void> {
-    const { client, state, logger, sessionId, messages, args } = ctx
+    const { client, state, logger, sessionId, messages, args, config } = ctx
 
     const params = getCurrentParams(state, messages, logger)
     const targetArg = args[0]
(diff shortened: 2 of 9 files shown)


ranxianglei:
Superseded by v1.11.0.

The `[ACP system message — not a user comment]` wrapper has been **removed entirely** in v1.11.0 (`lib/ui/notification.ts`), not just gated behind a debug flag. Notifications rely on `noReply: true` + `ignored: true` flags alone.

This PR is no longer needed. Recommend closing.

PR #97, opened 2026-07-10
feat: annotate message tags with token size + type

ranxianglei:
## Summary

Each `<dcp-message-id>` tag now includes `tokens` and `type` attributes, giving the model per-message size and type awareness without relying solely on ACP's breakdown recommendations.

**Before:**
```xml
<dcp-message-id>m00175</dcp-message-id>
```

**After:**
```xml
<dcp-message-id tokens="20.7K" type="tool:bash">m00175</dcp-message-id>
```

## Why

The model has no built-in token counter — it cannot tell which messages are consuming the most context. Previously, ACP compensated by injecting a breakdown listing the largest items. Now the model can scan the conversation itself and identify large messages independently.

## Changes

- **`lib/message-ids.ts`**: `formatTokenSize()` (e.g., 20700 → "21K") + `classifyMessageType()` (text/tool/reasoning + tool name)
- **`lib/messages/inject/inject.ts`**: `injectMessageIds` computes tokens + type per message via `countAllMessageTokens`
- **`lib/prompts/system.ts`**: ACP TAGS section updated to describe the annotation; breakdown de-emphasized as suggestion not directive
- **`tests/message-ids.test.ts`**: 12 new tests

## Verification

- TypeScript: 0 errors
- Build: dist/index.js 372K
- Deployed to local plugin cache

### DIFF
--- a/devlog/2026-07-10_message-token-sizes/REQ.md
+++ b/devlog/2026-07-10_message-token-sizes/REQ.md
@@ -1,0 +1,21 @@
+# REQ: Message Token Size + Type Annotation
+
+## Problem
+
+The model cannot see how many tokens each message consumes. ACP injects a breakdown
+listing the largest items, but the model has no per-message size awareness when
+scanning the conversation itself.
+
+## Solution
+
+Annotate each `<dcp-message-id>` tag with `tokens` (approximate size) and `type`
+(content classification: text/tool/reasoning + tool name).
+
+Before: `<dcp-message-id>m00175</dcp-message-id>`
+After:  `<dcp-message-id tokens="20.7K" type="tool:bash">m00175</dcp-message-id>`
+
+## Impact
+
+- Model can self-assess which messages are large without relying on ACP's recommendation list
+- Breakdown recommendations can be de-emphasized (suggestion, not directive)
+- Compression decisions become smarter — model combines content + size + type
--- a/lib/message-ids.ts
+++ b/lib/message-ids.ts
@@ -116,6 +116,38 @@ export function formatMessageIdTag(
     return `\n<${MESSAGE_ID_TAG_NAME}${serializedAttributes}>${ref}</${MESSAGE_ID_TAG_NAME}>`
 }
 
+export function formatTokenSize(tokens: number): string {
+    if (tokens < 1000) return String(tokens)
+    if (tokens < 10000) return `${(tokens / 1000).toFixed(1)}K`
+    return `${Math.round(tokens / 1000)}K`
+}
+
+export function classifyMessageType(parts: any[]): string {
+    let hasTool = false
+    let hasText = false
+    let hasReasoning = false
+    const toolNames: string[] = []
+
+    for (const part of parts) {
+        if (part.type === "tool") {
+            hasTool = true
+            if (typeof part.tool === "string" && !toolNames.includes(part.tool)) {
+                toolNames.push(part.tool)
+            }
+        } else if (part.type === "text") {
+            hasText = true
+        } else if (part.type === "reasoning") {
+            hasReasoning = true
+        }
+    }
+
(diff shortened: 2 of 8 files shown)


ranxianglei (commented):
## Review

Reviewed all 4 commits. Overall approach is sound — per-message token annotations are the right direction for letting the model self-assess. Found one performance concern and one type safety issue.

### 🔴 Performance: `countAllMessageTokens` uses the real tokenizer on every message every transform

`injectMessageIds` calls `countAllMessageTokens(message)` for every message on every `experimental.chat.messages.transform` hook run (i.e. every LLM call).

Call chain: `countAllMessageTokens` → `estimateTokensBatch` → `countTokens` → `anthropicCountTokens` (WASM sync tokenizer).

On a 500-message session, this adds **500 tokenizer calls per LLM call**. Individual calls are fast (~0.5-2ms each), but at scale this is 250ms-1s of overhead on every single message transform.

The existing `estimateContextComposition` function (which runs in the same pipeline, `lib/messages/inject/utils.ts`) uses `Math.round(text.length / 4)` — the cheap char/4 estimate. So there's an inconsistency:

- The nudge breakdown shows `m00175 = 20.7K` (char/4 estimate)
- The per-message tag on m00175 shows `tokens="21K"` (real tokenizer)

Two different counts for the same message. And the system prompt says:

> The token size is approximate — treat it as a relative guide, not an exact count.

If approximate is fine (and it is — the model just needs relative sizes to prioritize), then `char/4` is the right choice. Consider using `countMessageCharacters(msg)` (already exists, returns char count) with `/4` rounding, matching `estimateContextComposition`:

```ts
const msgTokens = Math.round(countMessageCharacters(message) / 4)
```

This eliminates the tokenizer calls entirely and makes the per-message tags consistent with the breakdown numbers.

### 🟡 Type safety: `classifyMessageType(parts: any[])`

```ts
export function classifyMessageType(parts: any[]): string {
```

Per AGENTS.md strict TypeScript rules, `any[]` should be typed. The SDK exports part types — consider importing and using them.

### ✅ No merge conflict with PR #98

PR #97 touches:
- `system.ts`: ACP TAGS section (line ~8-10)
- `inject.ts`: `injectMessageIds` function (line ~476+)

PR #98 touches:
- `system.ts`: CONTEXT BREAKDOWN section (lines 76-83)
- `inject.ts`: `injectCompressNudges` function (lines ~249-258)

Different sections, no conflict. The two PRs are complementary: #97 gives the model per-message sizes (so it can self-assess), #98 removes the size-based recommendation lists (so the model isn't directed by size).

### ✅ Other findings (no issues)

- `formatMessageIdTag` signature change is backward-compatible (`Record<string, string | undefined>`)
- `escapeXmlAttribute` properly escapes attribute values for XML safety
- Tag format is stable for unchanged messages (prefix cache safe)
- Test fix iteration (commits 2-4) was messy but final state is correct: regex assertions use `[^>]*`, string literals use exact tags
- 619 tests pass

### Verdict

The tokenizer performance concern should be addressed before merge (switch to char/4). The type safety issue is minor. Otherwise the PR is solid and is the right direction.

ranxianglei (commented):
Re-reviewed commit d53aa85. Both concerns from my previous review are fully addressed:

✅ **Performance fix**: countAllMessageTokens → Math.round(countMessageCharacters(message) / 4). Eliminates the WASM tokenizer calls entirely. countMessageCharacters iterates the same parts + extractToolContent path, just sums char lengths instead of tokenizing. Now consistent with estimateContextComposition's estimation method — the per-message tag and the breakdown will show the same numbers.

✅ **Type safety fix**: classifyMessageType(parts: any[]) → classifyMessageType(parts: WithParts["parts"]). Uses the proper indexed access type.

LGTM. Ship it.

PR #158, opened 2026-07-18
fix: suppress nudge when all content is protected

ranxianglei:
## Summary

When all context is protected (e.g., user puts `read` in `protectedTools` and all messages contain `read` calls), the nudge still fires uselessly — model receives "go compress" but has nothing to compress.

PR #147's `filterSuppressed` only handles the case where compressible ranges exist but are filtered out. This PR adds the missing case: `compressible.length === 0 && protected.length > 0` → all protected → suppress nudge.

Emergency override (98% context) still fires regardless.

## Fix

**`lib/messages/inject/inject.ts`**:
```typescript
const allProtected = contextRanges.compressible.length === 0 && contextRanges.protected.length > 0
const nothingToCompress = filterSuppressed || allProtected
const shouldInject = nudgeAllowed && (!nothingToCompress || emergencyOverride)
```

The `protected.length > 0` check distinguishes "all protected" from "no refs assigned yet" (both arrays empty — edge case, not suppression).

## Tests

- "nudge suppressed when all content is protected (nothing to compress)"
- "emergency override fires even when all content is protected"

752/752 tests pass.

### DIFF
--- a/devlog/2026-07-18_suppress-nudge-no-compressible/REQ.md
+++ b/devlog/2026-07-18_suppress-nudge-no-compressible/REQ.md
@@ -1,0 +1,25 @@
+# REQ: Suppress Nudge When All Content Is Protected
+
+## Problem
+
+When a user configures high-frequency tools (e.g., `read`) in `compress.protectedTools`,
+most or all context becomes non-compressible. The nudge system still fires based on total
+token growth (including protected content), creating false-positive nudges with nothing to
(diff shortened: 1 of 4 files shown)


ranxianglei:
Superseded by #159 — #159's branch was based on this PR and the squash merge included all of #158's commits + test fixes + Scenario A coverage. Closing in favor of #159 which is now merged to master.

Issue #176, opened 2026-07-22
[Bug]: `allowSubAgents: true` repeatedly rewrites historical task results and invalidates provider prefix cache because subagent result cache is cleared on session switches


kratky-pavel:
### Bug Description

When `experimental.allowSubAgents` is enabled and OpenCode runs multiple background subagents, ACP 1.13.2 repeatedly invalidates the provider prefix cache in the parent session.

In the affected parent session, prompt caching initially grew normally, but after background subagents started completing and ACP performed the first compression, `cache.read` became stuck at approximately **22,016 tokens** while uncached `input` continued to grow to tens of thousands of tokens per request.

The affected session processed:

- **33** parent/orchestrator model turns
- **14** `task` tool calls
- **3** `compress` tool calls
- **589,878** uncached input tokens
- **760,320** cached input tokens
- only approximately **56.3%** of total input served from cache

By comparison, controlled sequential sessions without background subagents achieved approximately **96.5–98.3%** cache utilization and rebuilt a growing cache prefix immediately after compression.

The installed ACP 1.13.2 source indicates the following mechanism:

1. `experimental.chat.messages.transform` calls `injectExtendedSubAgentResults(...)` for every transformed request when `allowSubAgents` is enabled.
2. `injectExtendedSubAgentResults(...)` walks historical completed `task` tool parts.
3. It fetches the current child-session messages and obtains the latest assistant text through `buildSubagentResultText(...)`.
4. `mergeSubagentResult(...)` replaces the existing body of the historical `<task_result>...</task_result>` block.
5. The expanded value is cached only in the in-memory `state.subAgentResultCache`, keyed by `callID`.
6. ACP creates only one mutable `SessionState` for the plugin instance.
7. Switching between parent and child sessions calls `resetSessionState(state)`, which executes `state.subAgentResultCache.clear()`.
8. `subAgentResultCache` is not included in `saveSessionState(...)`, so it is not restored when ACP returns to the parent session.
9. The next parent transform therefore fetches the child session again and may replace the same historical `<task_result>` with a different child-session tail.

This repeatedly changes an old message in the middle of the parent prompt. A prefix cache can then reuse only the stable content before that message, which matches the observed fixed cache boundary.

There is also a related concurrency risk: asynchronous parent and child message transforms share the same mutable `SessionState`. A transform can resume after an `await` after another session has reset and repopulated that shared state. This may also explain inconsistent state-save values observed during rapid parent/child switching.

### Expected Behavior

[...]

PR #180, opened 2026-07-22
fix: remove subagent history rewriting that stalled prefix cache

ranxianglei:
## Summary

Removes `injectExtendedSubAgentResults` — the code path that rewrote historical `<task_result>` tool outputs in the parent agent's message history on every transform run. This was the sole source of the provider prefix-cache stall observed with `experimental.allowSubAgents: true`.

## Problem

`experimental.allowSubAgents` bundled two behaviors:
1. **Primary (kept)**: Let ACP run inside subagent sessions.
2. **Secondary (removed)**: Rewrite the parent agent's historical `state="running"` `<task_result>` placeholder with the completed subagent result.

The rewrite path had a fatal flaw: `subAgentResultCache` was cleared on every parent↔child session switch (`resetSessionState`) and was never persisted. So each transform run re-fetched the subagent session and produced a new historical message body, invalidating the provider prefix cache at the rewrite point. Observed: cache hit rate ~56% (healthy: 96–98%), prefix frozen at ~22K tokens.

The rewrite was also **redundant**: OpenCode natively appends a `state="completed"` message with the full subagent result immediately after the `task` call completes.

## Changes

- **`lib/hooks.ts`** — drop `injectExtendedSubAgentResults` from pipeline + imports.
- **`lib/compress/protected-content.ts`** — `appendProtectedTools` drops `allowSubAgents` param and the subagent fetch+rewrite branch; uses `part.state.output` as-is.
- **`lib/compress/{range,message}.ts`** — update call sites.
- **`lib/messages/inject/subagent-results.ts`** — deleted (82 lines).
- **`lib/subagents/subagent-results.ts`** — deleted (74 lines); `lib/subagents/` directory removed.
- **`lib/messages/index.ts`** — drop re-export.
- **`lib/state/{types,state}.ts`** — drop `subAgentResultCache` field (never persisted, no format change).
- **10 test files** — drop `subAgentResultCache: new Map()` from mock state factories.
- **`tests/e2e-message-transform.test.ts`** — update pipeline comment.
- **`AGENTS.md`** — drop `subagents/` from §2.1 module map.

## Verification

- `npm run typecheck` — clean
- `npm run test` — **835/835 pass**
- `scripts/ci/check-pr.sh` — all checks pass

## Backward compatibility

No persisted-state format change (removed field was never serialized). `experimental.allowSubAgents` still enables ACP inside subagent sessions — only the parent-history rewriting is gone.

Refs: Gitea `dog/opencode-acp#176`, Gitea `dog/opencode-acp#20`.

Ultraworked with [Sisyphus](https://github.com/code-yeongyu/oh-my-opencode)


ranxianglei:
https://github.com/ranxianglei/opencode-acp/issues/176

PR #181, opened 2026-07-22
fix: guide models to correct quality-gate rejection recovery path

ranxianglei:
## Summary

Fixes the deadlock reported in dog/opencode-acp#33, where a non-adaptive model (glm-5.2) got stuck retrying a rejected compression 23× — re-billing ~320K tokens each turn — because the **correct recovery path was undiscoverable**.

When the blocking pre-commit quality gate rejects a `compress` call, the old guidance told the model to "rewrite a more complete summary" — **impossible for large ranges** (a 243K-token range needs ≥9725 chars to pass the 1% retention floor). The escape hatch (`acknowledgeRisk`) had no `.describe()` in the tool schema, the system prompt never mentioned rejections, and the rejection message never suggested splitting.

## Changes

| File | Change |
|------|--------|
| `lib/compress/range.ts`, `lib/compress/message.ts` | `acknowledgeRisk` gains a `.describe()` (was the only param without one) — escape hatch now visible in the tool schema |
| `lib/compress/quality-gate/rejection.ts` | `buildQualityRejectionError` branches on range size: **>50K tokens → SPLIT guidance** (break into smaller contiguous ranges), smaller ranges → denser-summary guidance; both paths mention `summaryMaxChars` + `acknowledgeRisk` |
| `lib/prompts/system.ts` | new **COMPRESSION REJECTION HANDLING** section: 3 recovery paths in priority order (split → denser/longer → acknowledgeRisk) + an anti-loop rule |
| `tests/quality-gate-enforcement.test.ts` | +2 tests: large range → split guidance; small range → denser-summary guidance |

Gate pass/fail logic, state schema, persisted format, and exported signatures are **unchanged**. Only guidance text changes.

## Why split guidance is gated on absolute size only

L1 floor is `summaryChars ≥ originalTokens × 4 × 1%`. At >50K tokens the dense minimum exceeds 2000 chars and grows linearly — impractical for one summary, so splitting is the right advice. A small range with a too-terse summary (which also has a high ratio) still has a sane dense-summary path, so ratio is intentionally **not** used as a trigger (it caused false "split" advice for small ranges).

## Verification

- `npm run typecheck` — clean
- `npm test` — **837 pass, 0 fail** (includes 2 new tests)
- `npm run build` — success; `dist/index.js` contains all three new strings
- Existing rejection tests (range/stats/acknowledgeRisk assertions) still pass

## Out of scope (follow-ups noted in devlog)

- Anti-deadlock counter (auto-downgrade after N consecutive rejections) — needs state schema migration.
- README/AGENTS.md doc fix: the quality gate is documented as "non-blocking" but `qualityGate.enabled` also activates this blocking pre-commit gate.

Refs dog/opencode-acp#33

### DIFF
--- a/lib/compress/quality-gate/rejection.ts
+++ b/lib/compress/quality-gate/rejection.ts
@@ -111,11 +111,3 @@ ${HOW_TO_COMPRESS_RULES}`
 
     return new Error(message)
 }
-
-export function buildPreemptiveAcknowledgeError(): Error {
-    return new Error(
-        'Parameter "acknowledgeRisk": true was provided, but no quality gate rejection is pending. ' +
-            "This parameter is only valid immediately after a compression was rejected by the quality gate. " +
-            "Remove it and try again.",
-    )
-}
--- a/lib/compress/quality-gate/rejection.ts
+++ b/lib/compress/quality-gate/rejection.ts
@@ -26,7 +26,6 @@ function computeStats(plan: RejectionPlanInfo): {
     originalTokens: number
     summaryChars: number
     ratio: string
-    ratioNum: number | null
     retentionPct: string
 } {
     let originalTokens = 0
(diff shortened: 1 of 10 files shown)


lib/compress/quality-gate/rejection.ts
```diff
@@ -26,17 +26,57 @@ function computeStats(plan: RejectionPlanInfo): {
...
     const retentionPct =
         originalTokens > 0 ? ((summaryChars / (originalTokens * 4)) * 100).toFixed(2) : "?"
-    return { originalTokens, summaryChars, ratio, retentionPct }
+    return { originalTokens, summaryChars, ratio, ratioNum, retentionPct }
+}
+
+// A range this large cannot be summarized densely enough in one pass: the L1
+// retention floor (1% of originalTokens*4 chars) would demand a multi-thousand
+// char summary. At this size, splitting is the correct recovery, not rewriting.
+const LARGE_RANGE_TOKENS = 50_000
```
ranxianglei, line 47: 写死的代码

PR #193, opened 2026-07-25
fix: inactive block decompress + acp_status visibility

ranxianglei:
## Summary

Two fixes for inactive (consumed/GC'd/user-decompressed) compression blocks, addressing [Gitea #20](http://m1.redoxos.org:3000/dog/opencode-acp/issues/20) feedback:

### 1. Allow decompress of inactive standalone blocks
`resolveBlockTarget` in `lib/compress/decompress.ts` previously rejected any block where `activeBlocks.length === 0` with *"not active. It may have already been decompressed."* — even for standalone inactive blocks that are safe to decompress. The model could see the block's compress call in context but couldn't decompress it.

**Fix**: Removed the "not active" rejection. Standalone inactive blocks (user-decompressed, GC'd, orphaned) can now be decompressed. The nested-redirect for consumed blocks (inside an active parent) is kept — decompressing a consumed child directly would restore 0 messages since the parent still claims them.

### 2. Show inactive/consumed blocks in acp_status
`scope:"compressed"` previously only iterated `activeBlockIds`, completely hiding consumed/GC'd blocks. The model couldn't discover blocks consumed by secondary compression.

**Fix**: Now shows all blocks from `blocksById` with an `[inactive]` marker on the metadata line (not the topic line, per user feedback). Adds a summary line when inactive blocks are present.

## Files
- `lib/compress/decompress.ts` — remove "not active" rejection
- `lib/compress/status.ts` — show all blocks, add inactive marker
- `tests/acp-status.test.ts` — 3 new tests
- `tests/decompress-logic.test.ts` — 2 new tests

## Verification
- typecheck: ✅
- tests: 851 pass (846 existing + 5 new), 0 fail
- build: ✅

### DIFF
--- a/devlog/2026-07-25_inactive-block-fixes/REQ.md
+++ b/devlog/2026-07-25_inactive-block-fixes/REQ.md
@@ -1,0 +1,30 @@
+# REQ: Inactive Block Fixes
+
+## Problem
+
+Two issues with inactive (consumed/GC'd/user-decompressed) compression blocks:
+
+1. **decompress rejects inactive blocks**: `resolveBlockTarget` in `lib/compress/decompress.ts`
+   rejects any block where `activeBlocks.length === 0` with "not active. It may have already been
+   decompressed." — even for standalone inactive blocks that are safe to decompress. The model sees
+   the block's compress call in context but cannot decompress it.
+
+2. **acp_status hides inactive blocks**: `scope:"compressed"` only iterates `activeBlockIds`,
+   completely hiding consumed/GC'd blocks. The model cannot discover blocks that were consumed by
+   secondary compression, making it impossible to reason about the full compression history.
+
+## Fix
+
+1. **decompress**: Remove the "not active" rejection in `resolveBlockTarget`. Keep the nested
+   redirect (consumed blocks inside an active parent still redirect to the parent). Standalone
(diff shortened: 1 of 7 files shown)


lib/commands/decompress.ts
```diff
@@ -149,15 +149,6 @@ export async function handleDecompressCommand(ctx: DecompressCommandContext): Pr
             )
             return
         }
-
-        await sendIgnoredMessage(
```
ranxianglei: 这里为什么要删除

ranxianglei: 删除的是 standalone inactive block 的**拒绝分支**：旧逻辑在 `activeBlocks.length === 0` 且无活跃父级时，发 "Compression X is not active." 通知并直接 return——这正是本 PR 要修掉的拒绝。删掉后这类 block 落到正常 decompress 流程；另外两个 guard（does not exist / 嵌套在活跃父级里提示先恢复父级）都保留了，安全性不受影响。

补充：该文件（`/acp decompress` 斜杠命令）已随 #233 整体删除，此 hunk 在 rebase（9f3a3ea）中已移除。等价修复在工具路径 `lib/compress/decompress.ts` 的 `resolveSingleBlockTarget`（删 "not active" error、保留 nested redirect）。

Issue #296, opened 2026-08-13
[Bug]: Error: __DCP_CONTEXT_HANDLED__


BsoBird:
### Bug Description

1. use opencode 1.18.18
2. opencode plugin opencode-acp@latest --global --force
3. use cmd /acp and get error.
4. check log.
```
timestamp=2026-08-13T05:03:29.343Z level=ERROR run=f8dd8df5 message=failed ref=err_22f418e7 error="Error: __DCP_CONTEXT_HANDLED__" cause="Error: __DCP_CONTEXT_HANDLED__\n    at <anonymous> (/home/bsobird/.cache/opencode/packages/opencode-acp@latest/node_modules/opencode-acp/dist/index.js:10690:17)\n    at processTicksAndRejections (native:7:39)\n    at Plugin.trigger (/$bunfs/root/chunk-zebeatt9.js:1098:1361)\n    at Plugin.trigger (definition) (/$bunfs/root/chunk-n6qzp1at.js:439:28058)\n    at SessionPrompt.command (/$bunfs/root/chunk-gzs24a6h.js:4:13932)\n    at SessionPrompt.command (definition) (/$bunfs/root/chunk-zebeatt9.js:1096:15462)\n    at SessionHttpApi.command (/$bunfs/root/chunk-gzs24a6h.js:2:84381)\n    at SessionHttpApi.command (definition) (/$bunfs/root/chunk-gzs24a6h.js:4:13842)\n    at Server.listen (/$bunfs/root/chunk-gzs24a6h.js:6:15456)\n    at Server.listen (definition) (/$bunfs/root/chunk-gzs24a6h.js:6:15579)"
```

did i miss something?
timestamp=2026-08-13T05:03:35.715Z level=INFO run=f8dd8df5 message=cleanup prune=7.days

### Expected Behavior

no error

### Debug Context Logs

```json
[...]

PR #297, opened 2026-08-13
fix: remove __DCP_CONTEXT_HANDLED__ throw that leaked to error logs (#296)

ranxianglei:
## Problem (Issue #296)

`/acp` commands throw `Error: __DCP_CONTEXT_HANDLED__` which leaks to the opencode error log. In opencode 1.18.18, this produces `level=ERROR` entries for every `/acp` invocation.

Reported by @BsoBird on opencode 1.18.18 + ACP 1.14.16.

## Root Cause

The command handler in `hooks.ts` threw `new Error("__DCP_CONTEXT_HANDLED__")` after handling `/acp` commands (lines 295, 299). This sentinel was meant to abort command processing, but opencode catches and logs it as an error.

## Fix

Replace both `throw new Error("__DCP_CONTEXT_HANDLED__")` with `return`. Commands deliver output via `sendIgnoredMessage` (writes directly to session via `client.session.prompt`), so the hook can return normally without side effects.

## Verification

- TypeScript: 0 errors
- Tests: 977 pass, 0 fail (added regression test)
- Build: 391.50 KB
- Deployed locally

### DIFF
--- a/devlog/2026-08-13_fix-dcp-context-handled-error/REQ.md
+++ b/devlog/2026-08-13_fix-dcp-context-handled-error/REQ.md
@@ -1,0 +1,13 @@
+# REQ — Fix __DCP_CONTEXT_HANDLED__ Error Leak (Issue #296)
+
+## Problem
+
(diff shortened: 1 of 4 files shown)


evilhikari:
Please merge this PR, thanks!

PR #313, opened 2026-08-16
fix: invalidate stale modelContextLimit on model switch (#312)

ranxianglei:
## Problem (issue #312)

切换模型导致上下文水平计算错误 — 配置了 50% 紧急压缩，但在 26% 时就被告急（20w 切到 100w）。

`state.modelContextLimit` is written **only** by the `system.transform` handler, which OpenCode fires **after** `messages.transform` in every turn (verified in opencode stable v1.14.45: `session/prompt.ts:1583` → `session/llm.ts:118`). On the first turn after a model switch, every percentage-based threshold is therefore computed against the **previous** model's window:

- `emergencyThresholdPercent: "50%"` × stale 200K = 100K → a 260K-token session (26% of the new 1M window) trips the false emergency
- same staleness distorts percentage `min/maxContextLimit` and the adaptive `nudgeGrowthTokens` (10K instead of 50K)

## Fix

Track **which model** the cached limit belongs to; invalidate it when the turn's model no longer matches:

- `SessionState.modelProviderID?` / `modelID?` — provenance of the cached limit (additive, persisted, `typeof`-guarded restore; old state files load unchanged)
- `syncModelIdentity()` (`lib/state/utils.ts`) — single invalidation point: on model mismatch (or unverifiable provenance from a legacy file) sets `modelContextLimit = undefined`, so the affected transform takes the existing, well-tested **"limit unknown"** path (no false emergency/min-max nudges; GC/truncation no-op; adaptive growth floor)
- `createSystemPromptHandler` records the identity alongside the limit; `createChatMessageTransformHandler` calls `syncModelIdentity` right after state resolution — before every limit consumer
- system.transform repopulates the correct limit + identity **later in the same turn**, so the degradation is exactly one transform (same semantics as a fresh session's first turn)

Deliberately **not** fetching limits via `client.app.providers()`: an extra API call + error/caching surface in the hot path, previously rejected as a design goal (per-session-state design doc).

## Verification

- `npm run typecheck`, `npm run build`, full suite: **990 pass / 0 fail** (981 baseline + 9 new)
- New `tests/model-switch-limit.test.ts`:
  - e2e #312 regression: 200K→1M switch, 260K tokens, 50% emergency → no false emergency before **or** after the system-hook refresh; **verified to FAIL with the fix disabled** (`stale 200K limit must be invalidated`)
  - control: 800K/1M at 50% still fires the emergency (mechanism not neutered); steady-state same-model keeps the limit
  - 5 `syncModelIdentity` unit tests + persistence round-trip
- Dual-agent review (AGENTS.md §5.3): both APPROVE, no Critical/Major; 2 P3 findings (printWidth on 3 new lines; fresh-session log wording) fixed in this commit
- Backward compatible: persisted format change is additive; older plugin versions ignore the new fields

Devlog: `devlog/2026-08-16_model-switch-stale-limit/` (REQ + DESIGN + WORKLOG)

### DIFF
--- a/devlog/2026-08-16_model-switch-stale-limit/DESIGN.md
+++ b/devlog/2026-08-16_model-switch-stale-limit/DESIGN.md
@@ -1,0 +1,75 @@
+# DESIGN - Invalidate stale modelContextLimit after a model switch
+
+- Task ID: `2026-08-16_model-switch-stale-limit`
+- Home Repo: `opencode-acp`
+- Created: 2026-08-16
+- Status: Accepted
+
+## 1. Problem Statement
+
+- **What problem are we solving?** Percentage-based context thresholds are computed against `state.modelContextLimit`, which lags the active model by one turn after a model switch (issue #312: 50% emergency fired at 26% of the new 1M window).
+- **Why now?** User-visible false emergency nudges on a common workflow (switching to a larger-window model to continue an existing session).
+
+## 2. Goals & Non-Goals
+
+- **Goals**:
+  - No threshold is ever computed from a limit whose model identity does not match the current turn.
+  - Self-correcting within one turn without extra API calls.
+  - Additive, backward-compatible persisted state.
+- **Non-Goals**:
+  - Perfect first-turn accuracy after a switch (accepted 1-turn "limit unknown" window).
+  - Upstream hook-ordering change in OpenCode.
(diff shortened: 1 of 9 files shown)


ranxianglei:
采纳了#314

PR #322, opened 2026-08-18
fix: stabilize system prompt token estimate (#255)

### LINKED ISSUE #255: estimateSystemPromptTokens inconsistent between nudge and acp_status after compression
## Bug

`estimateSystemPromptTokens` (introduced in PR #241) gives **wildly different results** depending on which call site invokes it:

- **Nudge breakdown** (`inject.ts`): reported **162.3K** system tokens (53% of context)
- **acp_status** (`status.ts`): reported **32.6K** system tokens (19% of context)

Same function, same session, 5× difference. The real system prompt is ~30-40K (AGENTS.md + tool definitions + ACP rules), so acp_status is closer to correct.

## Root Cause

`estimateSystemPromptTokens` scans the messages array for the first assistant message with token data, reads its `tokens.input + cache.read + cache.write`, and subtracts the first user message's token count:

```typescript
systemTokens = firstAssistantMsg.totalInput - firstUserMsgTokens
```

After compression replaces older messages, the "first visible assistant message" may be from turn 100+, where `totalInput` includes system + 100 turns of context — not just system + firstUser. So `systemTokens` is grossly inflated.

The two call sites see different message array states (nudge runs early in message-transform, acp_status is a later tool call), so they find different "first assistant messages" with different token data.

## Impact

- Nudge shows `system 53%` → model thinks system prompt dominates context → misleads compression strategy
- `compressibleTokens = total - systemTokens` is miscalculated → affects nudge gating
- Model sees inconsistent numbers between nudge and `acp_status` → confusion

## Reproduction

1. Start a long session with many compressions
2. Observe nudge breakdown system tokens vs `acp_status` system tokens
3. They diverge significantly

## Environment

- v1.14.8 (master `d087522`)
- Introduced in PR #241

## Fix Direction

Cache system prompt size once from the earliest API call (when context = system + first user message only). Store in `state.systemPromptTokens` and reuse. The system prompt doesn't change between turns unless the user edits AGENTS.md.

Beatrice0377:
## Summary

ACP already populated `state.systemPromptTokens` before pruning, but nudge and
`acp_status` ignored that cache — each re-estimated the system prompt tokens
from a different message view. After compression, the first visible assistant
can belong to a much later turn whose input includes large accumulated history,
inflating the estimate and causing the two numbers to diverge.

## Root Cause

- The stable pre-prune cache (`state.systemPromptTokens`) existed but was never
  consumed.
- nudge re-estimated from the post-prune transform array, where compressed
  messages are removed and the first visible assistant is a late turn.
- `acp_status` re-estimated from its own fetched message view.
- After compression those views can select different late assistant messages,
  producing inflated and divergent estimates.

## Fix

Three minimal production changes:

1. `cacheSystemPromptTokens` preserves the first stable positive estimate —
   later transforms no longer overwrite it with a degraded post-compression
   estimate (write-if-undefined guard).
2. `estimateContextComposition` prefers `state.systemPromptTokens` when it is a
   positive value; falls back to the existing
   `estimateSystemPromptTokens(messages)` unchanged when undefined.
3. `acp_status` (`collectVisibleMessages`) prefers the same cached value; falls
   back to the existing estimate when undefined.

## Scope

- No persistence change.
- No model/provider identity invalidation.
- No `/acp context` change.
- No public API change.
- No version bump.

## Tests

- Pure regression tests: degraded post-compression visible history must not
  inflate the system estimate (cached 10K survives first-visible-assistant
  input of ~200K), and undefined cache falls back to the original estimate.
- `acp_status` overview regression asserting the cached value appears in the
  breakdown.
- §5.7 multi-turn growth-cycle test (`preserveRecentMessages: 20`): baseline →
  growth → nudge → compress → new baseline → growth → nudge again, asserting the
  stable system estimate survives compression.
- Docker E2E: autonomous nudge → compress → growth → re-nudge with a verifier
  that ≥2 nudge observations report the same system prompt estimate.
- Bug-targeting tests verified to FAIL with the fix reverted and PASS with it
  applied.
- Verified: `npm run typecheck`, `npm run build`, `npm run test` (1010 pass),
  Docker E2E (12/12 scenarios), `check-pr.sh`.
- Live OpenCode smoke: plugin loads, ACP state initializes, `acp_status`
  executes without ACP errors. This verifies the runtime path only — the precise
  #255 regression is covered by the automated tests; the live environment was
  not used to reconstruct a 200K post-compression session.

## Known limitations

- The cache is in-session only; a restart after host compaction may re-estimate
  from degraded history.
- Mid-session instruction/AGENTS.md changes do not invalidate the cache.

Both are intentionally out of scope for this fix.

Fixes #255

### DIFF
--- a/devlog/2026-08-18_fix-system-prompt-tokens/DESIGN.md
+++ b/devlog/2026-08-18_fix-system-prompt-tokens/DESIGN.md
@@ -1,0 +1,114 @@
+# DESIGN - Stabilize system prompt token estimate (#255)
+
+- Task ID: `2026-08-18_fix-system-prompt-tokens`
+- Home Repo: `opencode-acp`
+- Created: 2026-08-18
+- Status: Accepted
+
+## 1. Problem Statement
+
+- **What problem are we solving?** After ACP compression removes the true first
+  assistant message from visible history, nudge and `acp_status` re-derive the
+  system prompt token estimate from the _first visible assistant_ — a much later
+  turn whose `input` includes large accumulated history. Both inflate the
+  estimate, and because each consumer sees a different message view (post-prune
+  transform array vs. fresh store fetch), the two numbers diverge.
+- **Why now?** Issue #255. The `SessionState.systemPromptTokens` field and its
+  writer (`cacheSystemPromptTokens`) already exist but have no readers — the
+  missing link is wiring consumers to the cache.
+
+## 2. Goals & Non-Goals
+
+- **Goals**:
+    - nudge and `acp_status` use the same stable system prompt token estimate
+      after compression changes visible messages.
+    - Preserve existing behavior exactly when no cache exists (`undefined`).
+- **Non-Goals** (per human review of the investigation phase):
+    - No persistence of `systemPromptTokens` (survives only in-session).
+    - No model/provider identity invalidation.
+    - No `/acp context` changes.
+    - No refactor of `lib/token-utils.ts`; `estimateSystemPromptTokens()` public
+      semantics unchanged.
+
+## 3. Current Architecture
+
+**How it works today:**
+
+```
+estimateSystemPromptTokens(messages)   # token-utils.ts
+  ↑ 现算 ×4（互不共享）
+nudge:    injectCompressNudges → estimateContextComposition → estimateSystemPromptTokens(messages)
+acp_status: collectVisibleMessages → estimateSystemPromptTokens(rawMessages)
+/acp stats: buildStatusReport（同 acp_status）
+/acp context: analyzeTokens（内联实现）
+
+cacheSystemPromptTokens(state, output.messages)  # hooks.ts:233，prune 前，每轮写入
+  → state.systemPromptTokens（存在，但无人读取）
+```
+
+**Pain points:**
+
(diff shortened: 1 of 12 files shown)


Beatrice0377:
## Dual-agent review

Independent review of the current PR diff is complete.

- Reviewer A: **APPROVE** — no blocking or non-blocking code findings.
- Reviewer B: **APPROVE** — no blocking or non-blocking code findings.
- Source changes, regression tests, §5.7 multi-turn coverage, and Docker E2E changes were reviewed independently.
- No additional code changes are requested from the reviews.

## CI note

All code-validation checks are passing:

- CI / build ✅
- CI / test (Node 22) ✅
- CI / test (Node 24) ✅
- CI / e2e ✅
- PR Checks / pr-validation ✅

The only failing required check is `PR Build Artifact / build-artifact`.

Its checkout, `npm ci`, and `npm run build` steps succeed; it fails specifically at `npm publish` with `ENEEDAUTH` because `NODE_AUTH_TOKEN` is empty for this fork PR. I have not modified the workflow, package version, or npm authentication to work around this.

Leaving the artifact check for maintainer-side handling.

PR #325, opened 2026-08-20
fix: effective compressible accounting stops phantom-range retry loops

ranxianglei:
## Problem

Issue #37 session `ses_7fb5cbc89d7ee8ky0AsO7Wrp7X` showed a repeated-compression loop ("频繁压缩连续压缩"):

1. **Phantom loop (floors 156-175, ~10 retries)**: nudge recommended `m00141–m00150` as "7.8K compressible". After 4 successful compressions (b23-b26), the remaining messages were the last user message (soft-filtered), 2 protected compress anchors, and 1 empty message — all pipeline-filtered. Every retry failed as phantom; the nudge kept recommending the same range → infinite loop.
2. **Min-size retry (floors 204-206)**: displayed "10.8K compressible" resolved to 3066 chars in the pipeline → `minCompressRange` rejection → retry.
3. **Squeezed context**: remaining "compressible" content was 108-token compress confirmations; the growth-gated nudge kept firing.

## Root Cause

`buildCompressibleRanges` counts raw visible-message tokens without applying the compress pipeline's soft filters (last-user, empty, protected zone). Display overstated compressible content 3.6×.

## Fix

- `CompressibleRange.effectiveTokens`: token sum of messages that survive the pipeline's soft filters
- **Config-derived floor** `resolveEffectiveFloor(config)`: `minCompressRange ÷ 4` (default 1250 tokens); 0 when `minCompressRange: 0` — the recommendation tracks the pipeline's real rejection threshold (fixes E2E scenario 06, which runs with `minCompressRange: 0`)
- Universal phantom guard: `effective > 0` always required (all-soft-filtered ranges are phantom bait regardless of floor)
- Honest display: "0.8K effective of 13.6K"
- `nothingToCompress` extended with `allBelowMin` → nudge silent on squeezed contexts (non-emergency)

## Verification

- Full suite green (typecheck clean, +15 new tests)
- §5.7.3: surgical behavioral revert makes both incident regression tests FAIL; fix restored → green
- E2E scenarios 06/09/10/11/12 pass locally (5/5) after the config-derived floor fix
- Dual-agent review: Oracle verified core mirroring correct (autonomous/multi-user/pruned cases); flagged the hardcoded-floor divergence → addressed in 5924c69. Explore verified no parser breakage, tests exercise real paths.

Issue #216, opened 2026-07-27
Bug: Nudge injection loop — anchored nudges injected even when recommendation filter says "nothing to compress", and failed `compress` calls never reset pending-nudge state


ranxianglei:
# Bug: Nudge injection loop — anchored nudges injected even when recommendation filter says "nothing to compress", and failed `compress` calls never reset pending-nudge state

## Environment

- opencode-acp: **1.14.3** (from `~/.cache/opencode/packages/opencode-acp@latest`)
- opencode: 1.14.43
- Model context limit: 260,000 tokens
- Relevant `acp.jsonc`:
  ```jsonc
  {
    "pruneNotification": "detailed",
    "compress": {
      "maxContextLimit": "55%",
      "minContextLimit": "30%",
      "showCompression": true,
      "permission": "allow",
      "mode": "range"
    },
    "debug": "true"
  }
  ```

## Summary

In a long tool-heavy session (281 messages, ~135 bash calls), the plugin injected compress nudges on **every assistant step** (4 injections within ~36 seconds: 14:22:27 / 14:22:37 / 14:22:55 / 14:23:03). The model attempted `compress` **40 times, of which 28 failed** (mostly rejected by the plugin's own protected-zone rule), and each failure made the next nudge come *faster*. The result is a self-reinforcing loop:

```
nudge injected → model tries to compress → rejected by protected-zone guard
→ pending-nudge state never cleared → nudge threshold halved
→ next nudge arrives even sooner → repeat
```

Three separate defects combine to produce this loop.

---

## Defect 1: Anchored nudge is injected BEFORE the "nothing to compress" check

All line numbers refer to the bundled `dist/index.js` of 1.14.3.

In `injectCompressNudges` (line 8474), the order of operations is:

```js
// line ~8624: anchored nudge injected as soon as nudgeAllowed is true
if (nudgeAllowed) {
  applyAnchoredNudges(state, config, messages, prompts, ...); // ← already pushed into suffix message
}

// line ~8646: recommendation ranges computed AFTERWARDS
const recommendedRanges = filterRecommendedRanges(unprotectedCompressible, ...);

// line ~8674: "nothing to compress" evaluated AFTERWARDS
const nothingToCompress = filterSuppressed || allProtected || allInProtectedZone;
let shouldInject = nudgeAllowed && (!nothingToCompress || emergencyOverride);
```

[...]

PR #326, opened 2026-08-20
fix: emergency + nothing-to-compress emits /compact notice, not compress demands — issue #216 residual

ranxianglei:
## Problem

Incident `ses_7fb5cbc8` (issue #37): user config set min/max context limits to 20%, making the session (55% actual) permanently emergency. `inject.ts` gated nudges with:

```typescript
shouldInjectNudge = nudgeAllowed && (!nothingToCompress || emergencyOverride)
```

The `emergencyOverride` bypass demanded "Context limit reached — compress now" every turn even when `nothingToCompress` was true — no valid targets existed, so the model retried phantom ranges and every failure re-armed the nudge (~12 failed compressions in the incident). This is the residual of **issue #216**'s loop surviving the v1.14.4 fix through the emergency path.

## Fix

- `emergencyNoTargets = emergencyOverride && nothingToCompress`
- Regular nudges: `shouldInjectNudge = nudgeAllowed && !nothingToCompress` (emergency bypass removed)
- Notice: `shouldInjectNotice = emergencyNoTargets && noticeCadenceMet` instead of compress instructions
- Cadence reuses `lastNudgeShownTokens` + `growthFloor` — no per-turn nagging
- Emergency WITH compressible ranges → strong alert unchanged

## Notice Text Evolution (3 revisions, evidence-driven)

**v1**: "/compact / new session / relax protections" — user-command steps
**v2** (user feedback "这种是不是export比较好"): export-first ordering
**v3** (commit `8baebd2`, incident ses_7fb5c607): **model-actionable framing**. A headless `opencode run` model read the command steps as things IT should execute, looped 18 turns on retry-compare-reply, and died at 233K/262K context overflow. Slash commands are user-only in every mode — the model's ONLY viable action is informing the user. Notice now reads:

```
🚨 Context is critically full (N% of limit) and there is nothing left that can be safely compressed.
Do NOT retry compress on the same ranges — they will keep failing.
You cannot execute user commands yourself. Act now via your reply/message tool:
- Inform the user that context is full and compression is exhausted
- Recommend they run /acp export (archives compression summaries to a file), then /compact or start a new session
- Alternatively, ask them to relax protected-tool / preserve-recent settings so compression becomes possible
Then stop retrying and await the user's response.
```

## Tests

- 4 new tests in `inject.test.ts`: notice content (incl. reply/message-tool framing), 3-turn cadence regression (§5.7.1), incident-shape phantom regression (sub-floor ranges), production-config preserve-recent zone (§5.7.1: `preserveRecentMessages: 20`)
- Replaced test asserting the old loop-driving behavior
- 2 assertions in `model-switch-limits.test.ts` updated
- Full suite: **1024 pass / 0 fail**; §5.7.3 revert-check verified

## Dual-Agent Review (both effectively APPROVE)

**Oracle** (3m20s): growthFloor math verified (22.5K @ 1M model); test config paths traced.

**Explore** (3v3m22s): all 4 emergency×compressibility combinations traced — no regression; all `lastNudgeShownTokens` consumers benign; "compress now" string only in gated paths; E2E `detectNudge` unaffected. Flagged §5.7.1 gap → **addressed in cae3e94**.

## Stacked On

**PR #325** — merge that first; this branch is based on its tip (5924c69).

## Config Note

User's global config also updated: 20%/20% → 80%/80% (matches repo defaults).

### DIFF
--- a/devlog/2026-08-20_effective-compressible-accounting/REQ.md
+++ b/devlog/2026-08-20_effective-compressible-accounting/REQ.md
@@ -1,0 +1,54 @@
+# REQ — Effective Compressible Accounting (Retry-Loop Fix)
+
+## Problem
+
+Issue #37 session `ses_7fb5cbc89d7ee8ky0AsO7Wrp7X` showed a repeated-compression loop:
+
+1. **Phantom loop (floors 156-175, ~10 retries)**: The nudge recommended range
+   `m00141–m00150` as "7.8K compressible | 3.4K protected". The model compressed it
+   successfully 4× (b23–b26), but the remaining uncompressed messages were: the last
+   user message (soft-filtered by `filterLastUserMessage`), 2 protected compress
+   anchors, and 1 empty message — all filtered by the pipeline. Every retry failed
+   with "all messages already compressed" (phantom), and the nudge kept recommending
+   the same range → infinite loop.
+
+2. **Min-size retry (floors 204-206)**: After a successful compression (b27), the
+   nudge immediately recommended `m00199–m00207` as "10.8K compressible". The actual
+   pipeline-compressible content was 3066 chars (protected zone + soft filters) →
+   rejected by `minCompressRange` (5000) → retry.
+
+3. **Squeezed-context pressure**: After the context was fully compressed, the
+   remaining "compressible" content was 108-token compress confirmations
+   (individually sub-minimum). The growth-gated nudge kept firing anyway.
+
+## Root Cause
+
+The recommendation engine (`buildCompressibleRanges`) counts raw visible-message
+tokens without applying the compress pipeline's soft filters (last-user-message,
+no-meaningful-content, protected zone). The display overstates what a compression
+would free by 3.6× in the incident, baiting the model into guaranteed-failed calls.
+
+## Fix
+
+1. `buildCompressibleRanges` now computes per-range `effectiveTokens`: the token sum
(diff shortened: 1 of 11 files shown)


ranxianglei:
## 📦 Built Plugin Artifact

**Branch:** `2026-08-20_emergency-nothing-compressible` (8baebd2)

### Option A — Install from npm PR tag (recommended)

```bash
opencode plugin opencode-acp@pr-326 --global
```

Each push to this PR publishes a new version under the `pr-326` npm tag.

### Option B — Install from GitHub

```bash
opencode plugin "github:ranxianglei/opencode-acp#2026-08-20_emergency-nothing-compressible" --global
```

### Option C — Download artifact

1. Download the artifact from the [Actions run](https://github.com/ranxianglei/opencode-acp/actions/runs/32383518850)
2. Extract the tarball and install:

```bash
tar xzf opencode-acp-pr326.tgz
cp -r package/dist ~/.cache/opencode/packages/opencode-acp@latest/node_modules/opencode-acp/dist
```

3. **Restart opencode** to pick up changes.

---
*This comment is automatically updated on each push.*

ranxianglei:
[system] 🔄 **ework-daemon** picked up this issue — preparing workspace…

ranxianglei:
[system] ❌ **ework-daemon** crashed (exit 143). Try posting again or @ework-daemon to retry.

PR #335, opened 2026-08-23
fix: self-disable when running under billion-context proxy

ranxianglei:
> **Model: qwen3.8-27b (vllm)**

## Problem

When both `opencode-acp` and the `billion-context` proxy MCP are loaded in opencode (via `bili opencode`), they register **identical tool names** (`compress`, `decompress`, `search_context`, `acp_status`). The proxy's tools should take priority.

## Fix

Add a 5-line guard after the `config.enabled` check: if `BILLION_CONTEXT_PROXY` is set (the launcher sets this env var), print a notice and return `{}` early — registering no tools. The proxy MCP then owns all ACP tools.

## Result

- `bili opencode` → opencode-acp disabled, proxy MCP active
- Standalone `opencode` → opencode-acp works normally (client-side compression)
- Both can coexist in plugin config without conflict

## Note

This is the opencode equivalent of billion-context-pi PR #211 (same pattern for pi).

PR #338, opened 2026-08-26
feat: self-disable in manual proxy mode (detect /bili/ in provider baseURL)

### LINKED ISSUE #337: Self-disable should also trigger in manual proxy mode (detect /bili/ in baseURL), not only via BILLION_CONTEXT_PROXY
## Problem

opencode-acp only self-disables when `BILLION_CONTEXT_PROXY` is set. That env var is set **exclusively** by the `bili <client>` launcher (e.g. `bili opencode`), which exports it into the spawned client process.

When a user runs **manual proxy mode** — `bili start` + pointing opencode's provider `baseURL` at the proxy (e.g. `http://127.0.0.1:8787/bili/https://<upstream>`) — the env var is absent, so opencode-acp stays active even though the proxy is in the request path.

**Result:** two competing ACP authorities — opencode-acp's client-side compression engine **and** the proxy's server-side compression engine (two uncoordinated states), duplicate tool names (`compress`/`decompress`/`search_context`/`acp_status`), and the `/acp` panel showing the wrong (client-side) state.

## Root cause

The guard in `index.ts` is env-var only:

```ts
if (process.env.BILLION_CONTEXT_PROXY) {
  console.log("[opencode-acp] disabled: BILLION_CONTEXT_PROXY detected — proxy handles compression")
  return {}
}
```

There is no detection of the proxy in the provider `baseURL`.

## Expected behavior

opencode-acp should **also** self-disable when its provider's `baseURL` contains the `/bili/` prefix (the zero-config proxy routing prefix). This matches the behavior already documented in billion-context `CONFIGURATION.md`:

> The `/bili/` prefix doubles as a **self-detection signal**: billion-context client extensions (billion-context-pi / opencode-acp) recognize it in their own baseUrl and self-disable, so you never get double compression.

So the documented self-detection is not actually implemented for the manual-proxy case — only the launcher (env-var) path works.

## Suggested fix

In addition to the `BILLION_CONTEXT_PROXY` env check, inspect the opencode provider config (available via the `config` hook / `ctx`) for any provider whose `baseURL` contains `/bili/`, and self-disable (log + `return {}`) when found. Keep the env-var check as-is for the launcher path; the two signals are OR-ed.

## Repro

1. `bili start --port 8787`
2. In opencode config, set a provider `baseURL` to `http://127.0.0.1:8787/bili/https://<upstream-host>`.
3. Keep `opencode-acp` in the `plugin` list.
4. Run plain `opencode` (NOT `bili opencode`).
5. Observe: no `[opencode-acp] disabled` log — the plugin stays active → double compression + wrong `/acp` state.

## References

- billion-context#265 (user report + discussion)
- billion-context `CONFIGURATION.md` (documented `/bili/` self-detection signal)
- opencode-acp PR #335 (the env-var-only self-disable that shipped in v1.14.25)

<sub>🤖 ework agent · vllm-qwen/qwen3.8-27b</sub>

---
_Mirrored from ework issue #337_

ranxianglei:
Closes #337

## Problem

The v1.14.25 self-disable (PR #335) only checks `BILLION_CONTEXT_PROXY`, which is set exclusively by the `bili <client>` launcher. In **manual proxy mode** (`bili start` + provider `baseURL` pointed at the proxy, e.g. `http://127.0.0.1:8787/bili/https://api.openai.com/v1`) the env var is unset, so ACP stayed fully active on top of a proxy that already manages context.

## Fix

Detection now also runs in the plugin `config` hook (the only non-deadlocking point that sees the merged provider config — verified empirically against a live opencode instance; awaiting `client.config.get()` in the factory deadlocks because the server listens after the factory resolves):

- **`lib/bili-proxy.ts`** (new, pure): `findBiliProxyProviders()` scans provider `options.baseURL` (defensive top-level fallback) for the documented `/bili/` path prefix. Lookalikes (`/bilix/`, `bilibili.com`, bare `/bili`) do not match.
- **`index.ts`**: on detection the config hook (which fires before the first LLM request)
  - sets `permission: "deny"` for all five ACP tools — verified this removes them from the LLM tool list,
  - skips `/acp` command registration and `primary_tools` wiring,
  - sets a factory-scoped flag that no-ops all five ACP hooks (system prompt, message transform, text complete, command, event) via a type-safe `guard()` wrapper — no signature changes in `lib/hooks.ts`.
  - The flag is **assigned, not latched**: a config reload without the proxy restores ACP.
- The `BILLION_CONTEXT_PROXY` env-var fast path is unchanged.

## Tests

15 new tests, full suite **1044/1044 pass**, typecheck + build clean:

- `tests/bili-proxy.test.ts` — 11 unit tests (marker, null/non-object input, `options` vs top-level precedence, multi-provider, non-string values, lookalike negatives)
- `tests/bili-proxy-integration.test.ts` — 4 integration tests through the real plugin factory with isolated XDG homes: deny + skip-wiring on detection; all hooks no-op (messages byte-identical, no system prompt appended); enabled path unchanged (command registered, `dcp-message-id` injection runs); re-enable after proxy removal

## Verification notes

- Live-opencode probe (opencode 1.14.46 + fake LLM server) confirmed: config hook receives merged provider config before first request; permission-deny removes plugin tools from the captured request.
- `scripts/ci/check-pr.sh` passes locally (branch name, devlog, version unchanged).
- Devlog: `devlog/2026-08-27_bili-proxy-self-disable/` (REQ / DESIGN / WORKLOG).

### DIFF
--- a/devlog/2026-08-27_bili-proxy-self-disable/DESIGN.md
+++ b/devlog/2026-08-27_bili-proxy-self-disable/DESIGN.md
@@ -1,0 +1,119 @@
+# DESIGN - Self-disable in manual proxy mode
+
+- Task ID: `2026-08-27_bili-proxy-self-disable`
+- Home Repo: `opencode-acp`
+- Created: 2026-08-27
+- Status: InProgress
+
+## 1. Problem
+
+The v1.14.25 self-disable (PR #335) only checks `process.env.BILLION_CONTEXT_PROXY`
+at plugin-factory time. That variable is set exclusively by the `bili <client>`
+launcher. Manual proxy mode (`bili start` + provider `baseURL` pointed at the
+proxy) leaves it unset, so ACP stays active on top of a proxy that already
+manages context.
+
+## 2. Where the provider config is observable
+
+Empirically verified against a live opencode 1.14.46 instance (probe harness in
+`.opencode/probe/`, logs `run*.log` / `fakellm.log`):
+
+1. **The plugin factory (`ctx`) does NOT expose the merged opencode config.**
+   `PluginInput` is `{ client, project, directory, worktree, experimental_workspace,
+   serverUrl, $ }`. Awaiting `ctx.client.config.get()` *inside* the factory
+   **deadlocks** (the HTTP server is not listening yet; the factory promise is
+   awaited before listen). Fire-and-forget succeeds ~30ms after the factory
+   returns — this is why `registry.hydrateModelLimitsFromClient` is already
+   fire-and-forget (index.ts).
+2. **The `config` hook** (`Hooks.config?: (input: Config) => Promise<void>`)
+   fires after the factory returns and **before the first LLM request**, and
+   receives the full merged `Config` including
+   `provider[name].options.baseURL`.
+3. **Setting `permission.<tool> = "deny"` for a plugin-registered tool in the
+   config hook removes that tool from the LLM's tool list** (captured request
+   no longer contained the probe tool). This is the verified lever for
+   tool-level disable.
+
+Therefore detection must happen in the `config` hook, not in the factory.
+
+## 3. Design
+
+### 3.1 Detection — `lib/bili-proxy.ts` (new, pure)
+
+```ts
+export const BILI_PROXY_MARKER = "/bili/"
+export interface BiliProxyMatch { provider: string; baseURL: string }
+export function findBiliProxyProviders(provider: unknown): BiliProxyMatch[]
+```
+
+- Scans every entry of the (possibly undefined) provider map.
+- Reads `options.baseURL` (the SDK-typed location); falls back to a top-level
+  `baseURL` defensively in case the config shape drifts.
+- Case-sensitive substring match on `/bili/`. Deliberately conservative: any
+  provider whose path contains the marker triggers the disable (sessions can
+  switch providers mid-run, so checking only the "default" model's provider
+  would miss switches). Lookalikes (`/bilix/`, `bilibili.com`, bare `/bili`)
+  do not match.
+- Pure function of its input → trivially unit-testable, no SDK imports.
+
+### 3.2 Disable wiring — `index.ts`
+
+Factory scope gains:
+
+```ts
+let disabledByBiliProxy = false
+const guard =
+    <TArgs extends unknown[]>(fn: (...args: TArgs) => Promise<void>) =>
+    (...args: TArgs): Promise<void> =>
+        disabledByBiliProxy ? Promise.resolve() : fn(...args)
+```
+
+- All five hooks (`experimental.chat.system.transform`,
+  `experimental.chat.messages.transform`, `experimental.text.complete`,
+  `command.execute.before`, `event`) are wrapped in `guard(...)`. The creators
+  in `lib/hooks.ts` are untouched — no signature changes, no new parameters
+  flowing through the hook pipeline.
+- The `config` hook, at its top:
+  1. `const biliMatches = findBiliProxyProviders(opencodeConfig.provider)`
+  2. `disabledByBiliProxy = biliMatches.length > 0` — **assigned, not
+     latched**, so a config reload without the proxy restores ACP (the flag
+     and the permission/config mutations are both re-derived per invocation).
+  3. If matches: log
+     `[opencode-acp] disabled: /bili/ proxy detected in provider baseURL (<names>) — proxy handles compression`
+     (console.log, mirroring the env-var guard's user-visible style), set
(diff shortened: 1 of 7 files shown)


ranxianglei:
[bot] Reviewing PR #338 now — branch `2026-08-27_bili-proxy-self-disable` (2 commits, 7 files, +850/−22). I'll check the diff, run the full test suite + typecheck + build locally, and verify the CI checks (branch name, devlog, version unchanged). Will report findings shortly.

<sub>🤖 ework agent · vllm-qwen/qwen3.8-27b</sub>

ranxianglei:
[bot] **Review complete — verdict: approve (no blockers).** Dual-agent review per AGENTS.md §5.3/§5.6.

**Local verification (all on branch `2026-08-27_bili-proxy-self-disable`):**
- `npm run typecheck` — clean
- `npm test` — **1044/1044 pass** (matches the issue claim)
- `npm run build` — clean
- `./scripts/ci/check-pr.sh 2026-08-27_bili-proxy-self-disable origin/master` — all checks pass (branch name, devlog REQ+WORKLOG, version unchanged at 1.14.25)
- Prettier clean on all PR-touched files (the repo-wide `format:check` failure of ~413 files is pre-existing on master, not from this PR)
- Devlog complete: REQ / DESIGN / WORKLOG present and accurate

**Code review findings:**

✅ **Correctness** — `findBiliProxyProviders()` (lib/bili-proxy.ts:48) handles null/undefined/non-object input, non-string/empty baseURL, and correctly rejects lookalikes (`/bilix/`, `bilibili.com`, bare `/bili`, `api.bili.example.com`). `options.baseURL` precedence over top-level is correct and tested.

✅ **Guard wrapper is type-safe and lossless** — verified all five handler signatures in lib/hooks.ts (createSystemPromptHandler:70, createChatMessageTransformHandler:139, createCommandExecuteHandler:330, createTextCompleteHandler:400, createEventHandler:409): all are `async` functions that mutate `output` in place and return `Promise<void>`, so the `guard()` wrapper drops no return values and propagates throws.

✅ **State integrity** — flag is assigned, not latched (index.ts:161); integration test 4 proves re-enable after proxy removal (deny → allow, ID injection resumes). The config-hook early return skips `hostPermissions` seeding, which is safe because the guard no-ops every reader of that state.

✅ **Type safety** — the `as any` on `messages.transform` is pre-existing on master (not introduced here); the `as typeof permission` cast is forced by incomplete SDK typings (SDK `Config.permission` has no index signature) and mirrors the pre-existing pattern. No `@ts-ignore`, no new assertions.

✅ **Tests** — both files import real source (detector + real factory via `await import("../index")`), XDG isolation set before import (correct — lib/config.ts captures env at module load), no tautologies, and the tests fail in both failure directions (guard-not-no-op → byte-identical assertion fails; flag-stuck → injection assertion fails).

**Non-blocking observations (optional follow-ups, no action required for merge):**
1. `console.log` in the config hook (index.ts:163) fires on every config reload while the proxy is present — style-consistent with the env-var guard, but `logger` is available if you want it quieter.
2. Tool removal relies on opencode honoring `permission: "deny"` for plugin tools in the LLM tool list — a cross-version host contract (verified against 1.14.46). If a future opencode changes this, tools would silently reappear. Worth a line in devlog follow-ups.
3. `includes("/bili/")` matches anywhere in the string (e.g. a query string `?redirect=/bili/` would trigger) — documented as an accepted trade-off in REQ/DESIGN, just noting the boundary.

**Per AGENTS.md §5.1.1.2, I will not merge the PR — that is a human-only operation.** CI checks (`pr-validation`, `test`, `build`) should be green; ready for you to merge whenever.

<sub>🤖 ework agent · vllm-qwen/qwen3.8-27b</sub>

ranxianglei:
## 📦 Built Plugin Artifact

**Branch:** `2026-08-27_bili-proxy-self-disable` (a48ee7a)

### Option A — Install from npm PR tag (recommended)

```bash
opencode plugin opencode-acp@pr-338 --global
```

Each push to this PR publishes a new version under the `pr-338` npm tag.

### Option B — Install from GitHub

```bash
opencode plugin "github:ranxianglei/opencode-acp#2026-08-27_bili-proxy-self-disable" --global
```

### Option C — Download artifact

1. Download the artifact from the [Actions run](https://github.com/ranxianglei/opencode-acp/actions/runs/32993699873)
2. Extract the tarball and install:

```bash
tar xzf opencode-acp-pr338.tgz
cp -r package/dist ~/.cache/opencode/packages/opencode-acp@latest/node_modules/opencode-acp/dist
```

3. **Restart opencode** to pick up changes.

---
*This comment is automatically updated on each push.*

PR #343, opened 2026-08-28
fix: gate T1 growth nudges on minContextLimit (issue #342)

### LINKED ISSUE #342: [Bug]: Growth nudges ignore minContextLimit and fire below it
## Bug description

ACP injects T1 compression nudges well below a configured model-specific `minContextLimit`.

I am using a 400K-context model with these settings:

```jsonc
{
  "compress": {
    "modelMinLimits": {
      "openai/gpt-5.6-sol": 150000
    },
    "modelMaxLimits": {
      "openai/gpt-5.6-sol": 215000
    },
    "nudgeGrowthTokens": 50000,
    "nudgeForce": "soft"
  }
}
```

The minimum is 150K, or 37.5% of the model window. During about two hours of normal use, ACP logged ten regular compression nudges. All ten had `trigger=growth`; none was caused by the 215K maximum. They fired at these context sizes:

```text
67,068
82,172
83,594
86,384
90,162
90,510
102,299
132,826
140,552
152,500
```

Some representative log entries:

```text
Compression nudge injected | trigger=growth currentTokens=67068 usagePct=16.8% growthSinceBaseline=66852 growthFloor=22500 recommendedRanges=1
Compression nudge injected | trigger=growth currentTokens=90510 usagePct=22.6% growthSinceBaseline=89710 growthFloor=22500 recommendedRanges=1
Compression nudge injected | trigger=growth currentTokens=102299 usagePct=25.6% growthSinceBaseline=60402 growthFloor=22500 recommendedRanges=3
Compression nudge injected | trigger=growth currentTokens=140552 usagePct=35.1% growthSinceBaseline=52564 growthFloor=22500 recommendedRanges=2
Compression nudge injected | trigger=growth currentTokens=152500 usagePct=38.1% growthSinceBaseline=54338 growthFloor=22500 recommendedRanges=2
```

The model normally follows the nudge and compresses old ranges, so its working context tends to stay around 80K to 150K instead of reaching the configured minimum.

## Expected behavior

I expected the limits to work like this:

```text
T1 efficiency nudge = currentTokens >= minContextLimit
                       AND growth >= nudgeGrowthTokens

T1 maximum nudge    = currentTokens >= maxContextLimit

T2/T3 nudge         = accumulated summary threshold
```

In this configuration, routine T1 compression would start after 150K, with stronger pressure at 215K. T2 and T3 consolidation should remain independent because it is useful even when the current working context is small.

## Source findings

The runtime behavior matches the current source:

- `computeShouldNudge()` uses `growthSinceLastNudge >= nudgeGrowthTokens || overMaxLimit`.
- `overMinLimit` selects the message variant but does not gate `shouldNudge`.
- `lib/messages/inject/utils.ts` marks `minNudgeContextPercent` as deprecated and ignored.
- `lib/messages/inject/inject.ts` passes `overMinLimit` into the decision but does not require it in `nudgeAllowed`.

This also means that raising `modelMinLimits` does not delay growth nudges. Raising `nudgeGrowthTokens` is not a clean workaround because the same setting controls T2 and T3 promotion.

Issue #300 covers percentage-based `nudgeGrowthTokens` and the broader decision chain. This report is narrower: a configured lower context limit does not act as a lower limit for T1 growth nudges. The explanation in issue #340 also says nudges require the context to exceed the minimum and pass the growth threshold, which does not match the current runtime behavior.

If bypassing `minContextLimit` is intentional, a separate T1 growth floor or a separate tier-promotion threshold would make this policy configurable without coupling unrelated behavior.

## Environment

- opencode-acp: 1.14.25
- context-compress-algorithms: 1.3.0
- OpenCode: 1.18.23
- Model: `openai/gpt-5.6-sol`
- Model context window detected by ACP: 400,000 tokens

ranxianglei:
## Summary

Fixes #342 — growth-triggered compression nudges fired well below a configured `minContextLimit`.

**Root cause**: `computeShouldNudge()` (external `context-compress-algorithms/trigger`) only uses `overMinLimit` to select the tips variant (`normal`/`minLimit`/`maxLimit`), not to gate the decision: `shouldNudge = growthSinceLastNudge >= nudgeGrowthTokens || overMaxLimit`. So a growth nudge fired whenever context grew past `nudgeGrowthTokens`, regardless of `minContextLimit`. This contradicted the README, which documents `minContextLimit` as the threshold below which reminder nudges are off. Turn/iteration anchors were already gated on `overMinLimit` — only the growth path bypassed it.

**Fix**: `lib/messages/inject/inject.ts` — add an explicit min-gate to the T1 growth path in `injectCompressNudges`:

```ts
const minGateOpen = !minLimitResolved || overMinLimit || overMaxLimit
const nudgeAllowed =
    emergencyOverride ||
    (decision.shouldNudge &&
        minGateOpen &&   // issue #342: min-gate
        growthSinceBaseline !== undefined &&
        growthSinceBaseline >= growthFloor)
```

- `overMaxLimit` implies `overMinLimit` in normal configs (`min < max`), so the max-limit path is unaffected; keeping `overMaxLimit ||` also makes the gate robust to a misconfigured `min > max` and to configs that set only `maxContextLimit`.
- **Resolvability guard** (added after dual-agent review): the gate only applies when `minContextLimit` resolves to a concrete value (`minLimitResolved`, newly returned by `isContextOverLimits` in `lib/messages/inject/utils.ts`). Models that don't report `limit.context` leave percent limits unresolvable — with a naive gate, `overMinLimit`/`overMaxLimit` would both be `false` and **every** growth nudge for such models would be suppressed (regression, verified live pre/post-fix). Unresolvable keeps pre-#342 growth-only behavior.
- The emergency override (default 98%) bypasses the gate.
- T2/T3 tier-promotion nudges are untouched (independent cadence, never consult `overMinLimit`).

The fix lives in ACP (not the shared `context-compress-algorithms` package) because `minContextLimit` semantics are ACP-specific policy. No dependency bump.

**Behavior note**: the default config sets `minContextLimit = maxContextLimit = 80%` (raised from 45% in v1.14.16 / PR #295), so for default users growth nudges now fire at ≥80% of the model context window (previously: as soon as growth crossed `nudgeGrowthTokens`). That is the documented intent of `minContextLimit`. Stale `45%`/`55%` defaults in `README.md` and `CONFIGURATION.md` are corrected as part of this fix.

## Tests

- **7 new tests** in `tests/inject.test.ts` (all prefixed `issue #342:`):
  - growth nudge suppressed below `minContextLimit`, fires once context crosses it (multi-turn, asserts `shouldInjectThisTurn` + `lastNudgeShownTokens` + `lastPerMessageNudgeTokens`)
  - full growth cycle: baseline → nudge → compress → new baseline → nudge (min-gate open)
  - min-gate holds in production config (`preserveRecentMessages: 2`)
  - over-max nudge bypasses the min gate (defensive `min > max` misconfig)
  - growth nudge still fires when `minContextLimit` is unresolvable (percent limit, unknown model limit) — regression lock for the resolvability guard
  - emergency override bypasses the min gate
  - T2 tier-promotion fires below `minContextLimit` (independent of the min gate)
- **4 existing tests updated**: `post-compress baseline then large growth DOES nudge`, `nudge threshold halves after first nudge (issue #23)`, `post-compress baseline then small growth does NOT re-nudge`, and `baseline initialized to currentTokens on first transform` set `minContextLimit` *above* the test context to keep it out of the way; the fix now correctly suppresses those, so their `minContextLimit` was lowered to keep them isolating the growth mechanism they test (the last two were caught by review — they passed for the wrong reason).
- Full suite: **1036 tests, 0 failures** (was 1029, +7). `npm run typecheck` clean, `npm run build` success.
- Verified per AGENTS.md §5.7.3 that the gate tests fail when the fix is reverted.

## Notes

- Dual-agent review (2 independent agents) found 1 major regression (naive gate suppresses all growth nudges for models that don't report `limit.context`), 2 masked existing tests, and missing acceptance-criteria coverage — all fixed in this PR.
- `npm run format:check` reports pre-existing repo-wide Prettier drift (the installed Prettier 3.9.5 reformats 412 files including unmodified `lib/config.ts`/`lib/hooks.ts`); CI does not enforce formatting. New code matches the file's existing style (4-space, no-semi, double-quote).
- No `version` bump (content commit). Devlog: `devlog/2026-08-28_min-gate-growth-nudges/`.


ranxianglei:

[bot] Review in progress. Initial status:

**Verified so far (all green):**
- Root cause confirmed in the external package: `shouldNudge = growthSinceLastNudge >= nudgeGrowthTokens || overMaxLimit` (`context-compress-algorithms/dist/chunk-BZYW3CH5.js:14`) — `overMinLimit` only selects the tips variant. README documents `minContextLimit` as "reminders are off" below it, so the fix aligns runtime with docs.
- `npm run typecheck` clean, `npm run build` success, full suite **1033/1033 pass**.
- Bug-detection check (AGENTS.md §5.7.3): with the fix reverted, the 2 gate-specific new tests fail exactly at the min-gate assertion; re-applied → all pass.
- T2/T3 tier-promotion independence confirmed by inspection (they key off `!shouldInject`, never `overMinLimit`).
- CI pre-merge checks (`scripts/ci/check-pr.sh`) all pass; devlog present.

**Dual-agent review (2 independent reviewers) found issues — I've verified each and am pushing fixes to the branch:**
1. **MAJOR (regression, live-verified):** for models that don't report `limit.context` (local models, catalog miss — `lib/hooks.ts:101,214`), percent limits (default 80%/80%) resolve to `undefined` → both flags false → the new gate suppresses **all** growth nudges. Pre-fix they fired on growth alone (verified: pre-fix `true`, post-fix `false`). Fix: only apply the min-gate when `minContextLimit` actually resolves to a number.
2. **MAJOR (test isolation):** `tests/inject.test.ts:318` ("post-compress baseline then small growth does NOT re-nudge") keeps `minContextLimit = 550K` above its 253K turn-2 context — post-fix it passes for the wrong reason (min gate, not the 3K-growth threshold). The sibling test was updated but this twin was missed. Same class at :448.
3. **MINOR:** REQ/WORKLOG state the default `minContextLimit` as 45% — it's been **80%** since v1.14.16 (PR #295, commit 03622c2). With default min=max=80% this PR changes default behavior (growth nudges now only fire at ≥80%) — intended, but must be documented.
4. **MINOR:** REQ §5 acceptance criteria "emergency override bypasses min gate" and "T2/T3 independence" have no locking tests.

Working on the fixes now — will follow up with the full review + verification results.

bernhardberger (commented):
Thanks — the explicit `overMinLimit` gate correctly addresses the reported below-minimum growth nudges, and the added regression coverage matches the original reproduction.

One adjacent edge case remains: `overMaxLimit` bypasses the new minimum gate, but it does not bypass `growthFloor`. `nudgeAllowed` still requires:

```ts
growthSinceBaseline >= growthFloor
```

even when `overMaxLimit === true`.

For example, with `nudgeGrowthTokens: 100000`, `minNudgeGrowthRatio: 0.45`, baseline 200K, and max 215K, reaching 216K would not inject a maximum nudge because 16K growth is below the 45K floor. It would wait until approximately 245K.

That differs from the stated policy “T1 maximum nudge = `currentTokens >= maxContextLimit`.” Could the max path bypass `growthFloor` as well, or could a regression test clarify that this delay is intentional? A focused case would use baseline 200K, current 216K, min 150K, and max 215K.

This does not invalidate the #342 minimum-gate fix, but it matters when `maxContextLimit` represents a provider pricing or input boundary.

ranxianglei:

[bot] Dual-agent review complete, all findings fixed, pushed to the branch. **PR is ready for human merge** (AGENTS.md §5.1.1.2: merges are human-only — I can't merge it myself).

## Review verdict

Two independent review agents both returned **REQUEST CHANGES** on the original 6-line fix. Every finding was independently verified and is now fixed:

1. **Major — regression for models that don't report `limit.context`** (verified live): with the naive `(overMaxLimit || overMinLimit)` gate, a model that doesn't report `limit.context` (lib/hooks.ts leaves `state.modelContextLimit` undefined on catalog miss) makes percent limits (default 80%/80%) unresolvable → both flags false → **every** growth nudge suppressed. Fixed with a resolvability guard: the gate only applies when `minContextLimit` actually resolves (`minLimitResolved`, now returned by `isContextOverLimits` in lib/messages/inject/utils.ts). Unresolvable → pre-#342 growth-only behavior.
2. **Major — existing test passing for the wrong reason**: `post-compress baseline then small growth does NOT re-nudge` (tests/inject.test.ts:318) kept `minContextLimit` 550K above its 253K context — the "no re-nudge" assertion was satisfied by the new min gate, not the 3K growth threshold it isolates. Fixed (550K→200K).
3. **Minor — second masked test**: `baseline initialized to currentTokens on first transform` (tests/inject.test.ts:448), min 200K vs 55K/58K contexts. Fixed (200K→50K).
4. **Minor — untested REQ §5 acceptance criteria**: added tests for the emergency override bypassing the min gate and for T2 tier-promotion firing below `minContextLimit` (independent cadence), plus a regression lock for the unresolvable-limit fallback and a baseline-untouched assertion in the over-max test.
5. **Minor — devlog/doc inaccuracy**: REQ/WORKLOG said the default `minContextLimit` was "45%" — it's actually **80%** (raised in v1.14.16 / PR #295, commit 03622c2). README.md and CONFIGURATION.md also carried stale 45%/55% defaults — corrected to 80%/80% and the min-gate semantics documented.

## Behavior note (intended, now documented)

Default config has `minContextLimit = maxContextLimit = 80%`, so for default users growth nudges now fire at ≥80% of the model context window (previously: as soon as growth crossed `nudgeGrowthTokens`). That is the documented intent of `minContextLimit` — the runtime was contradicting the docs, and now the docs agree too.

## Verification

- `npm run typecheck`: clean
- `npm run test`: **1036 tests, 0 failures** (was 1029; +7 new)
- `npm run build`: success
- AGENTS.md §5.7.3: confirmed the gate tests fail when the fix is reverted
- `./scripts/ci/check-pr.sh` (branch name / devlog / changelog rules): all pass
- No `version` bump (content commit)

## Commits (pushed to `2026-08-28_min-gate-growth-nudges`)

- `0f35414` — original fix
- `f634222` — review fixes (resolvability guard, unmasked tests, 3 new tests, doc defaults, devlog)
- `6eb1ef0` — record commit hash in WORKLOG

PR body updated to reflect all of the above. Please merge when ready: https://github.com/ranxianglei/opencode-acp/pull/343

ranxianglei:
review the update

bernhardberger:
Thanks for clearing up the distinction. We do need the per-model override you mentioned.

Our primary native OpenAI GPT-5.6 models have a 400K context and should start T1 growth nudges at 150K, so the global value would need to be 37.5%. We also use `openrouter/z-ai/glm-5.3`, which currently advertises a 1,048,576-token context through the OpenRouter API. Its configured floor is 200K. The same 37.5% value would delay its first growth nudge until 393,216 tokens, nearly twice the intended floor. Keeping the default 15% instead would put native OpenAI at 60K and reproduce the early nudges from #342.

Could ACP support an optional per-model growth floor, perhaps following the existing model-limit style?

```ts
modelMinNudgeLimits?: Record<string, number | `${number}%`>
```

A useful precedence order would be:

```text
modelMinNudgeLimits[provider/model]
minNudgeContextPercent × model context
existing growth-only behavior when model context is unknown
```

That keeps the simple 15% default and lets mixed-model installations use exact token or percentage floors. The existing `modelMinLimits` can keep its current turn/iteration meaning.

This does not need to block #343 if you would rather handle it in a follow-up. I mainly want to make sure the mixed-context limitation is recorded before the current fix ships.

ranxianglei:

[bot] **Review of the update (`ecfe7f0`, floor-based redesign): verified, with a few stale-doc fixes in progress.**

## Verification

- **No-op claim confirmed**: `minNudgeContextPercent` appears in the external `context-compress-algorithms` package only in its input type (`dist/trigger/types.d.ts:12`) — never read by the policy implementation. It was genuinely dormant, so activating it as the floor is the right call.
- `npm run typecheck`: clean. `npm run test`: **1035/1035 pass**. Build: success.
- §5.7.3 revert check: with the floor gate disabled, the 3 suppression tests fail (`suppressed below floor`, `production config`, `T2 independence` — the last one correctly, since T2 only runs when T1 didn't fire); restored → all green.
- Design: agreed. The dual-agent review had already flagged the 80%-default consequence in its "behavior note" — @dog's direction (use the dormant 15% knob, keep `minContextLimit` for turn/iteration reminders) is the correct resolution, and the WORKLOG revision history now records it properly.

## Findings from the update (fixing now)

The redesign left four stale references to the superseded minContextLimit gate, all from the previous revision:

1. `README.md:328-331` — still says "growth/turn/iteration reminders are off" below `minContextLimit` (growth is now floored by `minNudgeContextPercent`).
2. `CONFIGURATION.md:168` (`minContextLimit`) — same, plus the unresolvable-limit fallback clause now describes the wrong field.
3. `CONFIGURATION.md:192` (`minNudgeContextPercent`) — "before **any** nudges are shown" is wrong; it's the growth-nudge floor (turn/iteration are gated by `minContextLimit`), and the over-max/emergency bypasses + unknown-model-limit fallback aren't mentioned.
4. `lib/messages/inject/utils.ts:224` — the `@deprecated ... ignored` comment on `minNudgeContextPercent` (the exact comment issue #342 quoted) is now misleading: the field is active, computed in `inject.ts`. Also the test section header (`tests/inject.test.ts:1877`) still says "must respect minContextLimit".

Minor, not blocking: the dedicated emergency-bypass test was dropped (the emergency path remains covered by pre-existing tests, e.g. `tests/inject.test.ts:733,1194`), and the over-max test lost its baseline-untouched assertion.

## Per-model growth floor (@bernhardberger)

Recorded as **issue #344** — "feat: per-model growth-nudge floor (`modelMinNudgeLimits`) for mixed-context installs" — with your proposed API and precedence order (`modelMinNudgeLimits[provider/model]` → `minNudgeContextPercent × model context` → growth-only when the model window is unknown), plus your GPT-5.6/GLM-5.3 numbers as the motivating example. It does not block #343; the global floor ships first.

I'll push the doc/comment fixes shortly. Merge remains a human-only step: https://github.com/ranxianglei/opencode-acp/pull/343

ranxianglei:

[bot] Doc/comment fixes from my review of `ecfe7f0` are pushed (`b71e10b` + `022f4b4` hash record):

- `README.md` — `minContextLimit` = turn/iteration reminders only; growth nudges governed by `minNudgeContextPercent`
- `CONFIGURATION.md` — both entries re-pointed (incl. over-max/emergency bypasses + unknown-window fallback for the floor)
- `lib/messages/inject/utils.ts` — the stale `@deprecated … ignored` comment on `minNudgeContextPercent` corrected (field is active; floor computed in `inject.ts`)
- `tests/inject.test.ts` — section header re-pointed from "minContextLimit gate" to "minNudgeContextPercent floor"
- WORKLOG — `ecfe7f0` hash recorded, doc-fix row added

Re-verified after the edits: `npm run typecheck` clean, `npm run test` **1035/1035**, `npm run build` success. No code-logic changes — docs/comments/devlog only.

Summary state of the PR:
- Floor redesign (`ecfe7f0`) verified: no-op claim confirmed, revert check passes, design matches @dog's direction
- Per-model growth floor tracked in **issue #344** (does not block this PR)
- Branch `2026-08-28_min-gate-growth-nudges` is up to date; merge is a human-only step: https://github.com/ranxianglei/opencode-acp/pull/343

ranxianglei:

[bot] **Reviewed `0b92dd1` — LGTM.** ✅

- **Scope verified**: whitespace-only, single file (`README.md`, 5 lines) — fixes the 9→8 space indentation on the `minContextLimit` comment block in the default-config example. That mis-indent was introduced by my doc-alignment edit in `b71e10b`; thanks for catching it.
- **Final state checked**: the block now aligns consistently with its neighbors (README.md:324-336, all 8-space).
- **No build/test/typecheck needed**: Markdown-only change, provably inert for code paths.

Branch state: local = origin at `0b92dd1`, working tree clean. Full branch history now:

| Commit | What |
|--------|------|
| `0f35414` | original minContextLimit gate (superseded) |
| `f634222` | dual-agent review fixes (resolvability guard, tests, doc defaults) |
| `6eb1ef0` | WORKLOG hash record |
| `ecfe7f0` | floor redesign: `minNudgeContextPercent` (15%) as the growth-nudge floor |
| `b71e10b` | doc alignment with the floor redesign |
| `022f4b4` | WORKLOG hash record |
| `0b92dd1` | README indentation fix |

Nothing outstanding on my side — the PR is ready for the human merge step: https://github.com/ranxianglei/opencode-acp/pull/343

ranxianglei:
**Default floor lowered 15% → 5%** (maintainer decision, follow-up to the review discussion) — commits `f3a3fc8`+`2419871`.

**Why**: a 15% floor binds when `15% × window > baseline + 50K` (growth threshold), i.e. windows ≥ ~400–667K. On 1M-window models (the dominant large-window cohort), every compress cycle gets suppressed until 150K context usage: the typical working cycle (baseline 10–50K → +50K growth → nudge → max ~110K → compress → land 20–50K → repeat) never reaches 150K, so nudging/compression is starved and the steady-state working range shifts ~2× upward. Since compress resets `lastPerMessageNudgeTokens` to the post-compress size, the floor re-binds **every** cycle, not just the first. For that cohort this is bug-level.

**Why 5% is safe**: with the fixed 50K growth threshold, a 5% floor only binds when `5% × window > baseline + 50K` — ≥ ~1.2–2M windows for typical baselines — i.e. inert for essentially all real working cycles. Escape hatches unchanged: set it higher (15–30%) explicitly, or `0` disables the floor entirely.

Changes: `lib/config.ts` default, inject.ts policy passthrough + floor fallback, `dcp.schema.json` (+ stale description fixed), CONFIGURATION en/zh, new mutation-sensitive default-lock test (verified: reverts to failing if the fallback regresses to 15; suite 1036/1036, typecheck + build clean).

**Model**: Claude Sonnet 4.6

ranxianglei:

[bot] **Reviewed `f3a3fc8` + `2419871` (default floor 15% → 5%) — LGTM, with one minor zh-doc fix pushed (`f081b28`).**

## Math verified (independent re-derivation)

Floor binds when `P% × W > baseline + 50K` (growth threshold):

| Default | Bind condition | Windows (baseline 10–50K) |
|---------|---------------|---------------------------|
| 15% | `W > 6.67 × (b + 50K)` | ≥ ~400–667K ✓ matches commit msg |
| 5% | `W > 20 × (b + 50K)` | ≥ ~1.2–2M ✓ matches commit msg |

The cycle-starvation argument is sound: on a 1M window the typical cycle peaks ~110K < 150K floor, and since compress resets `lastPerMessageNudgeTokens` to the post-compress size, the 15% floor re-binds **every** cycle — not a one-time shift. 5% is inert for essentially all real cycles while `0` remains the escape hatch (verified: floor = 0 → `currentTokens >= 0` always true; schema `minimum: 0`).

## Implementation verified

- `lib/config.ts` default 15→5 ✓; `inject.ts` passthrough + floor fallback `?? 5` + rationale comment ✓
- `dcp.schema.json`: both the property default and the example block updated; stale "per-message nudges" description fixed ✓
- `buildConfig()` factory pins `minNudgeContextPercent: 15` (tests/inject.test.ts:34), so the existing floor tests stay decoupled from the default change — the two re-pinned comments are accurate ✓
- WORKLOG §7 amendment is thorough and matches the diff ✓

## Verification (ran myself)

- `npm run typecheck`: clean · `npm run build`: success · `npm run test`: **1036/1036**
- **Mutation check**: reverted the floor fallback to `?? 15` → exactly the new default-lock test fails (54/55 in the file); restored → clean. The test is genuinely mutation-sensitive, as claimed.

## Minor fix pushed (`f081b28`)

`CONFIGURATION.zh-CN.md` description still had the pre-redesign wording ("触发**任何** nudge 的最低上下文使用率百分比" — "any nudge"), which contradicts the actual behavior (only growth nudges; turn/iteration are governed by `minContextLimit`). The commit updated the zh default but not the description (pre-existing staleness from the floor redesign). Now mirrors the English entry, including the low-default rationale and `0`-disables note.

Branch is at `f081b28`, tree clean. PR ready for the human merge step: https://github.com/ranxianglei/opencode-acp/pull/343

Issue #346, opened 2026-08-28
[Bug] In-flight prune is a no-op (postTokens == prePruneTokens) at 229k tokens → request exceeds server context → empty-response resume loop


ranxianglei:
## Mechanism (proven from plugin daily logs)

Two production sessions (headless opencode, resumed per-message by an orchestrator) grew until every request failed. The plugin's own log shows the root cause:

```
2026-08-27T15:59:26Z Chat transform complete | session=ses_7fb333d7… messages=145 prePruneTokens=229479 postTokens=229479 nudged=false
2026-08-28T14:28:31Z Chat transform complete | session=ses_7fb2e8b9… messages=99  prePruneTokens=229535 postTokens=229535 nudged=false
```

**`postTokens == prePruneTokens` on every transform — even at 229k tokens.** The in-flight prune never removes anything, `nudged=false` throughout, and the persisted state shows `lastCompaction: 0` (proactive compaction also never fired; `stats.totalPruneTokens` 37k/230k respectively, all content parked in `prune.messages`).

## Failure chain

1. Conversation reaches ~229k tokens (plus ~15-20k system+tool schema + `max_tokens: 16384`)
2. Total exceeds the serving backend's context window (sglang qwen 27B, `max_model_len: 262144`) → **immediate request rejection**
3. opencode run ends with **exit 0 and zero output** (no error surfaced)
4. Orchestrator retries → identical ~4.6s failure, forever (3× consecutive observed per session)

Both sessions froze at almost exactly the same size (229,479 / 229,535) — the length-rejection wall.

## Environment

- opencode 1.14.46 + `opencode-acp@latest` (plugin log says v=1.14.25), direct-to-sglang (no bili in this path)
- Headless per-message spawn+resume (ACP path), daemon-driven
- `acp.jsonc` contains only the `$schema` ref (all defaults)

## Expected

The plugin's core contract is bounding the conversation. At minimum:
- In-flight prune must actually reduce tokens before the request when near the model context limit (or prune thresholds are misconfigured/never fire in this mode)
- Proactive compaction (`lastCompaction`) should trigger long before 229k
- On unrecoverable context overflow, surface an error instead of a silent empty run with exit 0 (the orchestrator can't distinguish this from success)

## Workaround in use

Orchestrator detects the empty-response loop and starts a fresh session (loses in-session continuity; conversation history only survives via the issue thread).

[...]

PR #349, opened 2026-08-28
fix: context-limit safety net for spawn+resume mode (#346)

ranxianglei:
## Root cause (issue #346)

In headless per-message spawn+resume mode the model context limit was **never known** when the messages-transform pipeline ran:

1. The system hook is the *only* writer of `state.modelContextLimit`, and it runs **after** `messages.transform` within a request — so on the (only) request a spawned process handles, the limit is still `undefined` during all threshold math.
2. The system hook **never persisted** the limit it learned (`saveSessionState` only runs inside the messages-transform pipeline) — the next spawned process started from `undefined` again. Learned and lost, every message, forever.
3. The init-time catalog seed (`hydrateModelLimitsFromClient`, fire-and-forget) races server readiness in spawned processes; nothing retried.

With `state.modelContextLimit === undefined`, every percentage threshold (`maxContextLimit`/`minContextLimit` = "80%", `emergencyThresholdPercent`, `gc.majorGcThresholdPercent` = "100%") resolved to `undefined`, and every consumer treated that as "do nothing" — nudges, batch-cleanup GC, and in-flight tool-output truncation were all silently disabled. Sessions grew to the length-rejection wall (229,479 / 229,535 tokens on a 262144 window + ~17K system + 16K `max_tokens`) → immediate rejection → exit-0 empty run → infinite retry loop.

A second, independent defect: even with a known limit, in-flight truncation started at **100% of the window** — already past the serving wall once system prompt + tool schemas + `max_tokens` are added.

## Changes

1. **Persist the limit** — the system hook now saves `modelContextLimit` + model identity on change, so a freshly spawned process resumes with the limit known. (State file shape is additive; the fields already existed.)
2. **Lazy catalog hydration** — on a catalog miss during `messages.transform`, the catalog is hydrated once per process from `client.config.providers()` (the server is guaranteed up during a request) before threshold math.
3. **`compress.contextLimitFallback`** (new, default `128000`, `0` disables) — when the limit is genuinely unknown, a configurable fallback window drives nudge thresholds, the emergency override, batch cleanup GC, and in-flight truncation. The real model limit always takes precedence. Documented in CONFIGURATION.md/zh + schema + validation.
4. **Overhead-aware in-flight truncation** — threshold is now `min(gc.majorGcThresholdPercent × limit, limit − systemPromptTokens − 16384)` (`OUTPUT_RESERVE_TOKENS`). With the production numbers (limit 262144, system ~17K, 229479 tokens) truncation now fires at 228760 instead of never. A window that cannot fit the overhead logs an ERROR and bails.
5. **Hard guard** — after the full transform pipeline, if post-transform tokens exceed `limit − systemPromptTokens − 16384`, ACP logs a loud `ERROR` ("ACP hard guard: …") with the budget breakdown. The exit-0 empty response itself is opencode-core behavior and cannot be changed from a plugin (upstream report candidate).

Also fixed along the way: internal-agent (title/summary/compaction) system prompts no longer overwrite the session limit with their own small model's limit.

## Testing

- **1049/1049 pass** (baseline 1029; +20 new tests across `tests/model-switch-limits.test.ts`, `tests/truncate-tools.test.ts`, new `tests/context-limit-fallback.test.ts`).
- **Pre-fix failure check**: with the source changes reverted, all 11 behavioral tests fail (lazy hydration, persistence, internal-agent guard, `hydrateAndResolve` ×3, hard guard, production wall repro, overhead bail, fallback truncation) — re-applying the fixes turns them green.
- **Production repro test**: limit 262144, `systemPromptTokens` 17000, currentTokens 229479 (the exact production token count) → truncation fires; pre-fix it was a no-op.
- All 10 pre-existing #312 model-switch tests pass unchanged (their configs carry no fallback, preserving legacy invalidation semantics).
- `npm run typecheck` + `npm run build` pass.

## Rollback / escape hatches

- `"compress": { "contextLimitFallback": 0 }` restores the exact legacy behavior.
- Larger `max_tokens` deployments: lower `gc.majorGcThresholdPercent` (the `min()` keeps the stricter bound).

Devlog: `devlog/2026-08-28_spawn-resume-context-limit/` (REQ + DESIGN + WORKLOG).

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -194,7 +194,7 @@ Core compression behavior.
 - **Type:** `number`
 - **Default:** `128000`
 - **Status:** ACTIVE
-- **Description:** Fallback context window (absolute tokens) used when the model's limit is unknown — e.g. custom providers with no declared limit, or headless spawn+resume sessions where the limit was never learned. Drives all percentage thresholds (`maxContextLimit`/`minContextLimit`), the emergency nudge override, batch cleanup GC, and in-flight tool-output truncation. The real model limit always takes precedence when known. Set to `0` to disable the fallback (legacy behavior: no safety net until the limit is learned).
+- **Description:** Fallback context window (absolute tokens) used when the model's limit is unknown — e.g. custom providers with no declared limit, headless spawn+resume sessions where the limit was never learned, or the brief window after a model switch invalidates a stale limit. Drives all percentage thresholds (`maxContextLimit`/`minContextLimit`), the emergency nudge override, batch cleanup GC, and in-flight tool-output truncation. The real model limit always takes precedence when known, as do per-model overrides (`modelMaxLimits`/`modelMinLimits`). Set to `0` to disable the fallback (legacy behavior: no safety net until the limit is learned).
 
 #### `compress.nudgeFrequency`
 - **Type:** `number`
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -190,6 +190,12 @@ Core compression behavior.
 - **Status:** DEPRECATED
 - **Description:** **Deprecated — scheduled for removal alongside `minContextLimit`.** Per-model override for `minContextLimit`. Still honored until removed.
 
+#### `compress.contextLimitFallback`
+- **Type:** `number`
+- **Default:** `128000`
+- **Status:** ACTIVE
+- **Description:** Fallback context window (absolute tokens) used when the model's limit is unknown — e.g. custom providers with no declared limit, or headless spawn+resume sessions where the limit was never learned. Drives all percentage thresholds (`maxContextLimit`/`minContextLimit`), the emergency nudge override, batch cleanup GC, and in-flight tool-output truncation. The real model limit always takes precedence when known. Set to `0` to disable the fallback (legacy behavior: no safety net until the limit is learned).
+
 #### `compress.nudgeFrequency`
 - **Type:** `number`
 - **Default:** `5`
--- a/CONFIGURATION.zh-CN.md
+++ b/CONFIGURATION.zh-CN.md
@@ -194,7 +194,7 @@ ACP 从最多三层配置文件中读取（后加载的覆盖先加载的）：
 - **类型：** `number`
 - **默认值：** `128000`
 - **状态：** ACTIVE
-- **说明：** 当模型上下文窗口未知时使用的回退窗口（绝对 token 数）——例如未声明 limit 的自定义 provider，或从未学到 limit 的 headless spawn+resume 会话。驱动所有百分比阈值（`maxContextLimit`/`minContextLimit`）、紧急 nudge 覆盖、批量清理 GC 与 in-flight 工具输出截断。已知真实模型 limit 时始终优先。设为 `0` 可禁用回退（旧行为：学到 limit 前无安全网）。
+- **说明：** 当模型上下文窗口未知时使用的回退窗口（绝对 token 数）——例如未声明 limit 的自定义 provider、从未学到 limit 的 headless spawn+resume 会话，或模型切换使旧 limit 失效后的短暂窗口。驱动所有百分比阈值（`maxContextLimit`/`minContextLimit`）、紧急 nudge 覆盖、批量清理 GC 与 in-flight 工具输出截断。已知真实模型 limit 时始终优先，按模型覆盖（`modelMaxLimits`/`modelMinLimits`）同样优先。设为 `0` 可禁用回退（旧行为：学到 limit 前无安全网）。
 
 #### `compress.nudgeFrequency`
 - **类型：** `number`
--- a/CONFIGURATION.zh-CN.md
+++ b/CONFIGURATION.zh-CN.md
@@ -190,6 +190,12 @@ ACP 从最多三层配置文件中读取（后加载的覆盖先加载的）：
 - **状态：** DEPRECATED
 - **说明：** **已废弃——将与 `minContextLimit` 一同移除。** 按模型覆盖 `minContextLimit`。在此之前仍然生效。
 
+#### `compress.contextLimitFallback`
+- **类型：** `number`
+- **默认值：** `128000`
+- **状态：** ACTIVE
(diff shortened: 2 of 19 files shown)


ranxianglei:
[bot] **Dual-agent review complete** (independent source + test reviewers, per AGENTS.md §5.3/§5.6). No blockers from either. Follow-up commit `24626ae` addresses the findings; suite is now **1050/1050** (was 1049, +1 §5.7 test), typecheck + build clean.

**Source (reviewer 1) — all minors, all fixed:**
- Threshold comment corrected: the `systemPromptTokens` term is *intentional* — provider-usage counts already include the system prompt (conservative margin there), while fallback counts exclude it (exact bound there); dropping the term would re-break the production case. Comment now says so.
- "Window too small" ERROR throttled to once per session (module-level Set).
- `hydrateAndResolve` now stores the in-flight promise instead of a boolean — concurrent callers share one hydration instead of one skipping it.
- System hook only writes the model identity pair when present — a limit without identity can no longer clobber the #312 staleness pair.
- `contextLimitFallback` docs extended (applies after switch-invalidation too; per-model `modelMaxLimits`/`modelMinLimits` take precedence) in config.ts, dcp.schema.json, CONFIGURATION.md/zh.
- Deferred (pre-existing, out of scope): non-atomic state-file write in `saveSessionState`.

**Tests (reviewer 2) — 2 majors, all fixed:**
- 7 pre-existing truncation tests were silently hollowed out by the earlier `makeState(1000)` → tiny-window bail (protection/skip branches never executed). Restored to realistic 200K windows; the branches now run.
- §5.7 gap closed: new multi-turn growth-cycle test (`growth cycle: fallback drives the nudge across turns…`) — fallback-only config, `preserveRecentMessages: 20`, three turns, asserting `shouldInjectThisTurn` + `lastPerMessageNudgeTokens` + anchor sets after every turn. The turn-2 anchor assertion is the pre-fix discriminator (pre-fix: `overMinLimit` never true → no anchors ever).
- Minors: poll for the persisted state file instead of a 100ms sleep; XDG env vars restored in `finally`; stub `hydrateAndResolve` mirrors once-per-process hydration; phantom `strategies` field removed from the new factory; faster verified fixture (MEDIUM_OUTPUT, 2201 tokens) for the two new truncation tests; stale threshold comments corrected.
- Deferred (pre-existing): config-factory gaps in older test files (tests are excluded from typecheck; runtime unaffected).

**Counting basis for the pre-fix failure check** (PR body said "11"): 16 tests assert the new behavior (19 if the fallback test file's import failure is counted); the "11" = the 11 named behavioral tests — the remaining 5 (hard-guard-silent, 2 fixed compat tests, etc.) pass by design pre-fix as controls.

Two verification bugs were caught and fixed in this round: `state.sessionId` is `string | null` (throttle key) and a missing `lib/messages/inject` barrel import in the new test file.

ranxianglei:

[bot] **Independent verification of final PR state (head `24626ae`) — all green.**

Re-ran the full verification suite on the branch after the follow-up review commit, and did a sanity pass over the diff:

**Local verification (this session, on `2026-08-28_spawn-resume-context-limit` @ `24626ae`):**
- `npm test`: **1050/1050 pass, 0 fail** (matches the post-review count in the review summary; the PR body's 1049 was pre-§5.7-test)
- `npm run typecheck`: clean
- `npm run build`: clean (dist/index.js 418 KB)

**CI (GitHub Actions on `24626ae`):** `pr-validation`, `test (22)`, `test (24)`, `build`, `e2e`, `build-artifact` — all **success**.

**PR state:** open, mergeable, `mergeable_state: clean`, not draft, 19 files / 2 commits, base `master`. `package.json` version untouched (no bump on a fix branch, per §5.1.1.1). Devlog `devlog/2026-08-28_spawn-resume-context-limit/` present with REQ + DESIGN + WORKLOG (DESIGN justified — data-flow changes across hooks/state/gc/inject).

**Sanity review of the diff — no issues found:**
- `hydrateAndResolve` (lib/state/state.ts:112-128) caches the in-flight *promise* — safe because `hydrateFromClient` (lib/state/model-limits.ts:47) has an internal try/catch and never rejects (returns 0 on failure), so there's no cached-rejection hazard for subsequent callers.
- Internal-agent guard ordering is correct: the `INTERNAL_AGENT_SIGNATURES` early-return (lib/hooks.ts:107) fires *before* the limit-attribution block (lib/hooks.ts:120), so title/summary/compaction agents on small models can no longer clobber the session limit.
- System hook writes `modelProviderID`/`modelID` only when present (lib/hooks.ts:134-138) — a limit without identity can't clobber the #312 staleness pair; `saveSessionState` fires only on actual change.
- `resolveEffectiveContextLimit` (lib/state/utils.ts:419-439): real model limit always takes precedence; fallback (`compress.contextLimitFallback`, default 128000) only when unknown; `undefined` when both unavailable — `contextLimitFallback: 0` restores exact legacy behavior, as documented.
- Overhead-aware truncation threshold `min(gc.majorGcThresholdPercent × limit, limit − systemPromptTokens − OUTPUT_RESERVE_TOKENS)` with the "window too small" ERROR throttled to once per session (module-level Set, lib/messages/truncate-tools.ts).
- Validation: `contextLimitFallback` must be a non-negative number (lib/config-validation.ts); schema + CONFIGURATION.md/zh-CN updated.

**Status:** dual-agent review (per §5.3/§5.6) was completed in the prior round with all findings addressed in `24626ae`; this round independently confirms the final state. The PR is ready — **merge is a human-only operation** (§5.1.1.2). Please merge when ready: https://github.com/ranxianglei/opencode-acp/pull/349

ranxianglei:

[bot] Re-checked the thread — this forward re-sent the dual-agent review summary (floor 3); there's no new request beyond what's already addressed, and the PR is unchanged since my last verification:

- Head still `24626ae` (2 commits), state open, `mergeable_state: clean`, not draft.
- Dual-agent review (§5.3/§5.6) complete, all findings fixed in `24626ae`.
- My independent re-verification (floor 6): 1050/1050 tests, typecheck + build clean, all CI checks green.

Nothing further needed from me. The PR is ready — **merge is a human-only operation** (§5.1.1.2). When you're ready: https://github.com/ranxianglei/opencode-acp/pull/349

Issue #344, opened 2026-08-28
feat: per-model growth-nudge floor (modelMinNudgeLimits) for mixed-context installs


ranxianglei:
## Requested by

@bernhardberger in the [PR #343](https://github.com/ranxianglei/opencode-acp/pull/343) thread (2026-08-28), as a follow-up to issue #342 / PR #343 (growth-nudge floor via `minNudgeContextPercent`).

## Problem

PR #343 makes the T1 growth-nudge floor the **global** `minNudgeContextPercent` (default 15% of the model context window). Mixed-model installations cannot express per-model floors with a single global percent:

| Model | Advertised window | Intended floor | Floor at global 37.5% | Floor at default 15% |
|-------|------------------|----------------|----------------------|---------------------|
| OpenAI GPT-5.6 (native) | 400,000 | 150,000 | 150,000 ✓ | 60,000 (reproduces #342 early nudges) |
| `openrouter/z-ai/glm-5.3` | 1,048,576 (OpenRouter API) | 200,000 | 393,216 (≈2× intended) | 157,286 |

No single global value satisfies both: 37.5% delays the GLM floor to nearly twice the intended value; 15% puts native OpenAI back at 60K.

## Proposed API

An optional per-model growth floor, following the existing `modelMaxLimits` / `modelMinLimits` style:

```ts
modelMinNudgeLimits?: Record<string, number | `${number}%`>
```

Keyed by `provider/model` (same keying as the existing model-limit overrides). Values may be absolute tokens or `"X%"` of that model's context window.

## Precedence order (as proposed by the requester)

```text
1. modelMinNudgeLimits[provider/model]
2. minNudgeContextPercent × model context
3. existing growth-only behavior when model context is unknown
```

This keeps the simple 15% default and lets mixed-model installations use exact token or percentage floors. The existing `modelMinLimits` keeps its current turn/iteration-reminder meaning (unchanged).

## Implementation notes

- Floor computation lives in `injectCompressNudges` (`minNudgeFloorTokens` / `overMinNudgeFloor`, `lib/messages/inject/inject.ts`).
- New config field needs validation/merging in `lib/config.ts` + `dcp.schema.json` + docs (`CONFIGURATION.md`, `README.md`), following the `modelMinLimits` pattern.
- Tests: extend the `issue #342:` block in `tests/inject.test.ts` (per-model floor wins over global percent; percent value resolves against the per-model window; unknown-model-context fallback unchanged).

## Scope

[...]

PR #370, opened 2026-09-08
feat: strip reasoning from protected-exempt historical messages (gated)

### LINKED ISSUE #368: Protected-message exemption inflates the incompressible context floor: per-compression reasoning is never reclaimable (~83.5% of measured residual)
## Summary

After repeated compression of a long session, 146 messages (~1.1 MB) remained that no compression block had ever covered. The largest component is `reasoning` parts (50.9%), and 83.5% of that reasoning is co-resident with `compress` tool parts. The proximate cause is that `filterProtectedToolMessages` excludes messages from compression selection at **message granularity** while `compress` is force-protected: the thinking emitted by each compression operation is thereby excluded from all future compression and re-sent with every request. The non-reclaimable context floor consequently grows monotonically with the number of compressions performed.

## Environment

- opencode-acp 1.14.27. All code-level statements below were verified against both the shipped `dist/index.js` and the TypeScript recovered from its embedded sourcemap (75/75 `lib/*.ts` files byte-identical).
- Observed session: ~1M-token context limit, 4,846 messages, 227 blocks created across the session's lifetime.

## Method

Read-only forensics over `opencode.db` (SQLite, `mode=ro`) joined against the plugin registry (`storage/plugin/acp/ses_*.json`). A message is classified as *uncovered* if it appears in no block's `effectiveMessageIds` and has no covering entry in `prune.messages.byMessageId`. Part size is `length(data)` in bytes.

## Measurements

Residual after the final compression (146 uncovered messages, 1,105,625 B):

| component | bytes | share |
|---|---|---|
| `reasoning` parts | 563,171 | 50.9% |
| `compress` tool parts (summaries in `state.input.content`, force-protected) | 153,939 | 13.9% |
| `skill` parts | 121,542 | 11.0% |
| remaining (other tools, text, step parts) | ~266,000 | ~24% |

Attribution of the 563,171 B of reasoning by co-resident tool part:

- **95.0%** (534,833 B) resides in messages containing at least one protected tool part
- **83.5%** (446,479 B) specifically in messages containing a `compress` part; session-wide, 120 compress-carrying messages hold 1,130,202 B of reasoning (mean 9,418 B, max 28,067 B per message)
- **0%** attributable to the recent-tail window (no uncovered post-last-compression messages without protected parts) — the residual is structural, not ordinary recent context

## Mechanism

1. `filterProtectedToolMessages` removes whole messages from every compression selection (`lib/compress/protected-content.ts:188-234`; `removedMessageIds.add(messageId)` — no part-level handling). A message's `reasoning` is therefore exempted together with its protected tool part.
2. `compress` and `skill` are the defaults for `compress.protectedTools` (`config.ts:158/245`), and `compress` is force-appended regardless of user configuration (`config.ts:167/476`).
3. Exempted messages never enter `byMessageId`, so the pruning survival rule (`lib/messages/prune.ts:60-89`) never removes them, and they are included in every outgoing request while the message is in the array.
4. Retention of the `compress` message itself is defensible: the live summary reaches the model only through that message (`range.ts:378`; `wrapCompressedSummary`, `compress/state.ts:55`). The `reasoning` part in the same message is not covered by that constraint — it never re-enters any selection (excluded prior to selection), and neither tier distillation nor `decompress` reads it (those operate on the registry and DB).

## Scope limitation

The figures above are DB-level totals accumulated over the session's entire lifetime, and bound the per-request residue from above; they do not represent the size of any single outgoing request, which was substantially smaller once consumed blocks were pruned and spliced. When a block is consumed by a higher tier, `hideConsumedCompressCalls` drops the `compress` part, and messages left with only structural parts are spliced (`compress/parts.ts:1-5` classifies `reasoning` as structural). Two subclasses have no such reclaim path: (a) the live-block window — every compression round adds an exempt message that persists until T2/T3 consumption; (b) `skill`-carrying messages — `hideConsumedCompressCalls` handles only `compress` parts. Exact per-request decomposition would require simulating the full pipeline and was not performed.

## Related findings

1. The displayed residual is estimated over text+tool only (`estimateContextComposition`, `inject/utils.ts:623-666`, total at `:703`; `length/4` heuristic) and excludes reasoning; `lastPerMessageNudgeTokens` is a lagging snapshot. The reported percentage therefore omits the largest measured residual component.
2. `byMessageId` entries whose `activeBlockIds` drain to empty remain permanently visible: `prune.ts:62` treats "entry present, no active blocks" as survive. Observed on 8 messages / 28,338 B.
3. `rewriteCompressInput` bails out conservatively when all batch siblings are consumed (`hide-consumed.ts:42`, `kept.length === 0 → return null`), leaving unmatched batch summaries in the request (~3.7 KB/block in this session).

## Possible mitigation

Strip `reasoning` parts at request time for messages that survive solely via the protected-message exemption, excluding the last unclosed assistant turn. Providers mandate replaying thinking blocks only for the current unresolved tool round (Anthropic thinking-signature validation; Gemini `thought_signature` on tool calls), so gating on turn-closure state rather than provider identity covers all providers. No persistent writes required.

ranxianglei:
## Problem

On long sessions, `reasoning` parts that ride on `compress`/`skill`-carrying assistant messages form a **permanently-incompressible context floor**. Protection is message-granular (`lib/compress/protected-content.ts:202`), so the whole message — reasoning included — is excluded from every compression and re-sent on every turn. Each compression round adds ~9 KB of unreclaimable reasoning; measured **~83.5% of the never-covered residual** in a real long session (#368). It is a monotonic feedback loop that degrades long-session usability.

## Solution

A single **request-time** pass, `stripProtectedReasoning`, that drops the `reasoning` parts from protected-exempt messages in **closed historical turns** before the request is sent. The current (possibly-open) round is always preserved. No persisted-state or internal-tag changes; no DB writes.

### Five gates (request-level gates 4–5 run first; all must hold to strip)
1. **Provider allowlist** *(added by review, 2026-09-08)* — `compress.stripProtectedReasoningProviders` (default `["anthropic","gemini"]`, `"*"` = all, case-insensitive substring). **Fail-closed**: unknown or undefined provider → nothing is stripped. Rationale: GPT-family gateways may reject requests whose historical thinking blocks are incomplete.
2. **Session activation** *(added by review)* — `compress.stripProtectedReasoningMinMessages` (default `100`, `0` = always). Below it the pass is a byte-stable no-op, so short sessions never pay prefix-cache churn.
3. **Turn-closure** — message index strictly `< lastUserIndex` (index of `getLastUserMessage`). The active round is never touched.
4. **Selector** — the message contains a tool part whose `part.tool` ∈ `config.compress.protectedTools` (`compress`/`skill`).
5. **Size** — total reasoning length > `threshold` (default **0** — review found per-message size is cache-noise: invalidation propagates from the first stripped message, so the activation gate is the real cache lever).

**Action**: `msg.parts = parts.filter(p => p.type !== "reasoning")` — drops reasoning only; the tool call + other parts are preserved (the live summary lives in the compress-call body, not the reasoning).

### Placement
Inserted **after** `hideConsumedCompressCalls` and **before** `assignMessageRefs` in `lib/hooks.ts` — operates on the minimal surviving set (reasoning-only leftovers are already spliced there, since reasoning is "structural" per `lib/compress/parts.ts:1`).

### Config (additive, safe defaults)
- `compress.stripProtectedReasoning` (bool, default `true`) — kill-switch.
- `compress.stripProtectedReasoningThreshold` (number, default `0`) — size threshold in chars.
- `compress.stripProtectedReasoningProviders` (string[], default `["anthropic","gemini"]`; `"*"` = all; explicit `[]` = strip for no provider) — **added by review**.
- `compress.stripProtectedReasoningMinMessages` (number, default `100`; `0` = always) — **added by review**.
- All registered in `VALID_CONFIG_KEYS` + `validateConfigTypes`; excluded from `CompressOverridableConfig` (global-only).

## Files
- `lib/messages/reasoning-strip.ts` — the pass, now with 5 gates + optional `options` param (omitted = ungated for pure-function callers).
- `lib/hooks.ts` — passes `state.modelProviderID` + config gates into the call.
- `lib/config.ts` — 4 config keys (interface + defaults + merge + `CompressOverridableConfig` exclusion).
- `lib/config-validation.ts` — `VALID_CONFIG_KEYS` + `validateConfigTypes` (incl. string[] element validation).
- `dcp.schema.json` — schema properties + defaults.
- `tests/reasoning-strip.test.ts` — 18 pass tests + 5 merge tests + 14 gate tests.
- `tests/e2e-message-transform.test.ts` — hook-level kill-switch, provider-gate (fail-closed), and activation-gate tests.
- `devlog/2026-09-07_strip-protected-reasoning/` — REQ / DESIGN / WORKLOG (review updates recorded).

## Testing
- **1124/1124 tests pass** (was 1096 at PR open; +28 across the review rounds). Typecheck clean, build clean.
- **Mutation-verified (§5.7.3)**: disabling the provider gate → 5 tests fail; disabling the activation gate → 2 tests fail; dropping the hook options object → 2 e2e tests fail; dropping the hook `??` fallbacks → 1 e2e test fails (each verified by temporarily mutating the code, then restoring).

## Review
- Original implementation: dual-agent review, both REQUEST-CHANGES; all findings addressed.
- 2026-09-08 review session (issue #368 re-review of this PR): recommended the provider allowlist + activation gate + threshold-0 changes now in this PR; owner approved direct application ("直接修改pr").
- Second dual-agent review of the additions (source + test focus): both **APPROVE**, all MINOR/NIT findings addressed — request-level provider resolution (`requestModel?.providerID ?? state.modelProviderID`, fixes undefined-provider on a fresh session's first request), allowlist entry trimming, 9 new config-validation tests (incl. fractional `minMessages` now rejected — `Number.isInteger`), hook-fallback e2e test (mutation-verified), tight boundary + presence assertions. One pre-existing follow-up noted: e2e `buildConfig` is structurally incomplete (tests are excluded from typecheck).

## Secondary findings (filed as separate issues)
- **A** — display/`acp_status` estimator excludes reasoning (`lib/messages/inject/utils.ts:586`) → #371.
- **B** — orphaned `byMessageId` entries with emptied `activeBlockIds` stay visible (`lib/messages/prune.ts:60-66`) → #372.
- **C** — `rewriteCompressInput` full-consumption leak (`lib/compress/hide-consumed.ts:42`) → #373.

Fixes #368


ranxianglei:
Superseded: owner redesign — the allowlist/activation-gate approach is replaced by a clean reimplementation: nested `compress.reasoning { drop, threshold }` config with the existing three-level provider/model cascade (#344), selector narrowed to `compress` tool calls only, single-thinking-size threshold. New PR incoming (refs #368).

Issue #422, opened 2026-09-17
V2: repeated valid system messages reject the entire context projection


drexb-ops:
## Scope

Dual-runtime fork `drexb-ops/opencode-acp@1fe36e0`, OpenCode `v2.0.3` (`d44b52c`).

## Reproduction

Lower two valid projected `system` messages with distinct IDs and identical text using official `toLLMMessages`. OpenCode emits two `Message.system(text)` values without IDs. Pass the projected and lowered arrays to ACP `normalizeV2ProjectedHistory`.

Actual: `valid: false`, `System source s1 has multiple lowered origins`.

The same diagnostic repeated in a real long-lived session immediately after activation of the #418 fix. Therefore the subsequent tool failures were not established to be solely a plugin-reload problem.

## Root cause

`lib/v2/projection/normalize.ts:661-673` correlates system sources by text and rejects more than one identical unclaimed candidate. Official OpenCode `packages/core/src/session/runner/to-llm-message.ts` lowers system messages without preserving their IDs. Repeated instruction text is a valid host sequence.

## Impact / fix direction

One repeated system text rejects the complete ACP context projection, skips pruning/ID/nudge injection, and prevents fresh initialization from committing. Use a verified ordered occurrence mapping where possible; otherwise preserve ambiguous system objects as opaque while processing independently correlated messages. Do not guess ownership or remove system messages.

Acceptance: genuine host-lowered repeated identical system messages, interleaved histories, extra host-added system messages, and rejection of truly ambiguous destructive edits. Assert a fresh registry commits and valid direct compression affects the following request.

Related: #418, #419, #395.

ranxianglei:
继续

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework) · qwen3.8-27b

**状态 / Status** — ✅ #422 的原始请求已完成并验证(修复 + 测试 + devlog,提交 `ec3ada23`);仅剩“开 PR”这一步因权限受阻,需要你处理。详见下方。

## Fix complete & verified

[...]

PR #429, opened 2026-09-17
fix(v2): map repeated identical system text by ordered occurrence

ranxianglei:
Fixes #422 — repeated identical projected system messages were rejecting the entire V2 context projection.

## Root cause
`lib/v2/projection/normalize.ts` correlated projected system sources to their lowered outgoing messages **by text** and rejected the whole projection when a single system text had more than one unclaimed candidate. OpenCode's lowering emits system messages **without preserving IDs**, so N identical projected records legitimately produce N identical ID-less lowered messages. One repeated instruction text therefore rejected the *entire* projection — skipping pruning / message-ID / nudge injection and blocking a fresh session from ever committing state.

## Fix
Correlate by **ordered occurrence** instead of rejecting: reuse the existing greedy `outgoingSystem()` claiming helper so the k-th projected record maps to the k-th unclaimed lowered match. Unmatched records stay opaque/unmapped; ownership is never guessed and no system message is dropped. All non-system rejections are unchanged (the fail-closed guard for genuinely ambiguous destructive edits is preserved).

## Tests
- 5 unit tests in `tests/v2-context-patch.test.ts` + 1 e2e in `tests/v2-context.test.ts` covering every acceptance criterion: repeated identical text, interleaved histories, extra host-added system messages, rejection of truly ambiguous patchable sources (regression guard), and fresh-registry-commit + direct-compression-affects-next-request.
- Key tests verified to **fail on the reverted buggy code**, then pass with the fix.
- `tsc --noEmit` clean · `npm run build` success · full suite green except a pre-existing environment-only failure in `tests/soft-block.test.ts` (hardcoded temp dir not writable in this sandbox; unrelated, not in the diff, passes on CI).

Devlog: `devlog/2026-09-17_v2-repeated-system-text/{REQ,WORKLOG}.md`.

## Note on target / base branch
This fix belongs to the V2 dual-runtime line that currently lives in `drexb-ops/opencode-acp` (sibling fixes #2/#3 landed there). This session's credential (`ranxianglei`) is read-only on `drexb-ops`, so I opened it here on `ranxianglei/opencode-acp` (where the tracking issue lives) against a **base branch pinned to the V2 tip** (`2026-09-17_v2-base` = `1fe36e09`) to keep the diff a clean 5-file change rather than dragging in the whole `lib/v2/` tree. If you'd prefer it land on `drexb-ops/opencode-acp`, grant `ranxianglei` push there (or pull branch `2026-09-17_v2-repeated-system-text` / commit `ec3ada23`) and I'll re-open it there.


ranxianglei:

[bot] 🏷 On it — this is flagged as an agent PR (`ework-agent-pr`), so I'm not re-doing the work. Reviewing now: pulling branch `2026-09-17_v2-repeated-system-text` against base `2026-09-17_v2-base` (1fe36e09), reading the diff in `lib/v2/projection/normalize.ts`, re-running the test suite locally, and checking CI status. Will report findings here shortly.

ranxianglei:

[bot] 🏷

**PR review complete — approved, ready for human merge** (after two small test-depth nits I fixed directly on the branch, commit `47e020f4`).

## What I verified

**Code** (`lib/v2/projection/normalize.ts`, ~21 lines):
- The fix removes only the system-specific multi-candidate rejection (`System source <id> has multiple lowered origins`) and routes system sources through the existing greedy `outgoingSystem()` helper (`lib/v2/projection/normalize.ts:106-117`). Drafts are processed in source order and per-text index lists are built in ascending lowering order, so k-th projected → k-th unclaimed lowered match holds by induction; interleaved texts get independent per-text occurrence counters.
- Unclaimed lowered messages land **provider-owned**: `opaque: true`, `owned: false`, `opaqueMessage = message` (`normalize.ts:866-884`). System entries are additionally `opaque/protected/allowSourceRemoval: false`, and `patch.ts` skips opaque entries from all edit paths — so even a hypothetical mis-attribution between identical twins cannot drive destructive patching.
- All non-system fail-closed guards untouched (duplicate IDs, role ambiguity, missing exact match for patchable sources, `ambiguous-origin` guards in `patch.ts`). No persisted-state schema change, no `dcp-*` tag change, no V1 path touched. Strictly less work than before (a per-draft `.filter()` was deleted).

**Tests** (diff review + independent revert-run):
- Reproduced your revert claim: with `normalize.ts` restored to the base version, 3 of the 5 new unit tests fail (`maps repeated identical…`, `preserves per-text occurrence…`, `keeps extra host-added…`); the other 2 are behavior-preservation/regression guards that pass on both sides, as expected. With the fix: 47/47 green across both v2 test files.
- Full suite on the PR head: **1416/1417 pass**. The single failure is `tests/soft-block.test.ts` — `EACCES: permission denied, mkdir '/tmp/…'` at module load (line 14), i.e. this sandbox's read-only `/tmp`; unrelated to the diff, passes on CI. Confirmed your claim.
- `tsc --noEmit` clean · build success · prettier clean after my nit commit.

**CI**: no GitHub Actions runs exist for this PR — both `ci.yml` and `pr-checks.yml` trigger only on PRs targeting `master`, and this one targets the pinned base `2026-09-17_v2-base`. The local verification above stands in for CI. (Branch-name/devlog checks would pass; changelog check N/A — `version` untouched.)

**Independent second-agent review** (per §5.3): verdict **APPROVE-WITH-NITS**, no blockers/majors. Key confirmation: mapping semantics are byte-identical to pre-fix except the deleted rejection branch, and system provenance is structurally non-editable downstream. I fixed both nits directly on the branch in `47e020f4`:
1. The "unclaimed" half of the two host-added tests was asserted only indirectly (absence of rejection) — now asserts directly that the extra lowered message lands provider-owned (`opaque`, `owned=false`, `sourceMessageId=undefined`, `opaqueMessage` identity).
2. Noted in the e2e that the `m99999` compress boundaries are deliberately bogus (reachability probe only) so nobody "fixes" them.

Plus a missing prettier wrap on one long line from the original commit. Format/typecheck/tests re-run after the push: all green. Head is now `47e020f4`.

(Environment note: the `oracle` subagent model returned `ProviderModelNotFoundError` and the `delegate_task` channel failed with `JSON Parse error: Unexpected EOF` even on trivial prompts, so the second review ran via a general-purpose agent instead.)

## Decision needed: target repo

My recommendation is explicit: **land this on `drexb-ops/opencode-acp`**, where the rest of the V2 line (#2/#3) lives. Simplest path: pull branch `2026-09-17_v2-repeated-system-text` (commits `ec3ada23` + `47e020f4`) into drexb-ops and open it there against the V2 tip, then close this PR here without merging — it stays as the clean-diff record. Merging here onto the pinned base would strand the fix off the active V2 line. I won't merge either way — PR merges are human-only.

---
中文摘要:修复了 V2 投影中重复相同文本的 system 消息导致整个投影被拒、新会话无法提交状态的问题(改为按出现顺序做第 k→第 k 个有序关联,fail-closed 守卫全部保留);已独立复验(回退验证新测试确实失败、全量测试 1416/1417 唯一失败为本沙箱只读 /tmp 的环境问题),并直接在分支上补了两处测试断言 nit(`47e020f4`),代码本身可以合并——但建议最终落在 `drexb-ops/opencode-acp` 而不是本仓的 pinned base。

PR #403, opened 2026-09-15
fix: normalize single Windows path separators in protectedFilePatterns

ranxianglei:
## Summary

Fixes #402 — ported from upstream DCP commit `5f8f33b`.

`normalizePath()` in `lib/protected-patterns.ts` did `replaceAll("\\\\", "/")`. In source, `"\\\\"` is the **two-character** string `\\`, so only doubled backslashes were ever replaced. Real Windows paths contain single separators, making the normalization a no-op on the only platform it exists for — `protectedFilePatterns` silently failed to protect Windows paths at all four `isFilePathProtected` call sites (protected-content, sweep, deduplication, purge-errors). Only patterns like `**/*.ts` accidentally worked, because `*` compiles to `[^/]*`, which spans backslashes freely.

## Changes

- `lib/protected-patterns.ts:2` — replace each single backslash with `/` (both input path and pattern are normalized identically, so cross-style matching stays consistent and `*` still compiles to `[^/]*` without crossing directory boundaries)
- `tests/protected-patterns.test.ts` (NEW) — 8 regression tests: Windows path vs forward-slash pattern, pattern written with Windows separators, wildcard boundary semantics, read/multiedit/apply_patch parameter shapes, unmatched-path negatives, `isToolNameProtected` non-regression, regex metacharacter escaping. Backslashes are built via `String.fromCharCode(92)` so no future reformatting can reintroduce an escape mistake.

## Verification

- New suite: **8/8 pass** on fixed code; **5/8 fail** against pre-fix code (verified by temporarily stashing the fix)
- Typecheck: clean (`tsc --noEmit`)
- Full suite: **1271 tests, 0 failures**
- Prettier: changed files clean (repo-wide `format:check` reports 452 pre-existing unformatted files unrelated to this change; left untouched to keep the diff minimal)

No state-format, config-schema, or API changes. No version bump (content fix).

PR #401, opened 2026-09-15
release: v1.18.1 — paper preprint v0.2 into the repo (docs only)

ranxianglei:
## v1.18.1 — paper preprint v0.2 into the repo (docs only)

Patch release after v1.18.0. Content = PR #394 (already on master); this PR is pure bookkeeping: version bump + changelog entries + devlog. **No runtime code changes.**

### Release content (already on master via #394)
- **Paper**: `paper/model-driven-incremental-hierarchical-compression-training-free-multi-generational-context-management-for-long-lived-coding-agents.md` (EN) + `paper/模型驱动的增量分层压缩-免训练多代上下文管理.md` (ZH) — preprint v0.2: production-scale longitudinal study (4.5 months, three hosts, 174,327 model calls, 18.76B cumulative input tokens, ~24.7B across all hosts, zero window violations on 204,800-token models, marathon sessions of 8,584–12,049 calls).
- **Figures**: `paper/figures/fig1-fold-layout.png` … `fig6-capacity-sim.png` (+ generator `make-figures.py`).
- Both READMEs carry an MIT open-access callout linking to their own-language version.

### Verification
- `./scripts/ci/check-pr.sh`: all green (branch name / devlog / changelog version string)
- `npm run typecheck` ✓ · `npm run build` ✓ · `npm run test`: **1263 pass, 0 fail**
- Note: `npm run format:check` fails repo-wide on pristine master too (452 files, pre-existing; not part of CI) — left untouched to keep the diff clean.

### After merge (automated)
Merge triggers `release.yml`: tag `v1.18.1` → npm publish (`latest`) → GitHub Release.

---

**中文摘要**：v1.18.1 补丁版本 —— 内容为 PR #394 论文预印本 v0.2 入库（纯文档，无运行时代码变更）；本 PR 仅含版本号 bump、中英 CHANGELOG 条目与 devlog。本地验证全绿（check-pr.sh / typecheck / build / test 1263 项全部通过）。请人工合并 —— 合并后 CI 自动打 tag、发布 npm latest 并创建 GitHub Release。

PR #390, opened 2026-09-11
fix(ci): skip npm publish on fork PRs (ENEEDAUTH) — keep artifact + comment

ranxianglei:
Fixes #366.

Fork PRs can't access `NPM_TOKEN` → `npm publish` in `build-artifact` fails with ENEEDAUTH and the whole job (tarball, artifact upload, install comment) goes red — e.g. PR #341.

- Publish step now gated: `if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}`\n- Comment's npm-tag section (Option A) hidden for forks, replaced by a note; Options B (GitHub) / C (artifact) unchanged\n- Fork PRs now get a green build + installable tarball; same-repo PRs unchanged

Verified: YAML parses, embedded JS `node --check` clean, upload glob still matches the base-version tarball when publish is skipped. CI-only change, no runtime code.

PR #389, opened 2026-09-11
fix: qualityGate.algorithms nested params no longer flagged as unknown keys

ranxianglei:
Fixes #329.

`qualityGate.algorithms.<algoName>.<param>` is legal config (consumed by `lib/compress/quality-gate/evaluate.ts`) but `getConfigKeyPaths` recursed into the dynamic map and every nested key failed the static allow-list → false `ACP: config warning — Unknown keys` on every startup.

Fix: add `qualityGate.algorithms` to the recursion skip list, same convention as `compress.providers` / `messageFilters.filters` (dynamic-key maps validated by their owning subsystem).

Tests: 2 new (nested params clean; unknown siblings still flagged). Suite 1209/1209; typecheck clean.

PR #433, opened 2026-09-17
promote: stable v1.18.1 — sync stable tag with latest

### LINKED ISSUE #346: [Bug] In-flight prune is a no-op (postTokens == prePruneTokens) at 229k tokens → request exceeds server context → empty-response resume loop

### LINKED ISSUE #347: [Bug] Silent context-overflow death on resume when model reports limit.context=0 (custom providers): uncalibrated thresholds → 400 → empty exit-0 loop

### LINKED ISSUE #368: Protected-message exemption inflates the incompressible context floor: per-compression reasoning is never reclaimable (~83.5% of measured residual)

### LINKED ISSUE #379: 支持压缩内容自定义存储位置

### LINKED ISSUE #384: perf: avoid transform work that scales with compression history
## Problem\n\nOpenCode becomes slow in long ACP sessions, especially after many compressions. A local isolated probe measured adaptive candidate planning at approximately 3.2 ms for 100 messages, 34 ms for 500, and 105 ms for 1,000.\n\n## Root causes\n\n- Candidate validation rebuilds global boundary and active-summary lookup structures for each draft.\n- T1 nudge analysis calculates full context/range data even when no nudge can be emitted.\n- Stable transforms replay historical inactive compression blocks and rebuild immutable consumed-call indexes.\n- Asynchronous state saves need ordered coalescing to prevent redundant writes and stale overwrites.\n\n## Proposed fix\n\nUse request-scoped search indexes, lazy no-nudge analysis, verified-state synchronization, transient derived indexes, and ordered state-save coalescing. Preserve candidate executor validation, protection semantics, decompression, fork recovery, and persisted-state compatibility.\n\n## Acceptance\n\nTarget candidate planning at 1,000 messages of 20 ms or less, with benchmark evidence and full regression coverage.

ranxianglei:
Promotes npm dist-tag `stable` from **1.14.26** → **1.18.1**, bringing `@stable` in line with the current `@latest` release. Requested in #400 (floor 6): "本项目先发一个stable版本".

## Why

- v1.18.1 has been running as `latest` since 2026-09-15 (PR #401, docs-only patch) — ~2 days of production validation
- `stable` has sat at 1.14.26 since 2026-09-03 (PR #358) — seven releases behind
- The README install command is `opencode plugin opencode-acp@stable --global`, so `@stable` users are missing high-severity fixes: context-limit safety net (#346) + budget guard (#347) in v1.17.0, and the transform performance fix 13.6 s → 1.2 ms (#384/#385) in v1.17.1

## What this PR does

No code changes — devlog only (`devlog/2026-09-17_promote-stable-v1.18.1/REQ.md` + `WORKLOG.md`). On merge, `release.yml` Pattern 3/4 detects the promote-stable branch name / commit title and runs `npm dist-tag add opencode-acp@1.18.1 stable`. `package.json` version is untouched → no changelog entry required (check-pr skips it).

## Content shipped in stable 1.14.26 → 1.18.1

| Version | Headline |
|---|---|
| v1.14.27 | Self-disable also triggers in manual proxy mode (`/bili/` baseURL detection, #338); soft deprecation of `minContextLimit`/`modelMinLimits` (#352) |
| v1.15.0 | `compress.reasoning` — drop oversized thinking from closed-turn compress calls (#377, fixes #368) |
| v1.16.0 | Top-level `storagePath` — relocate session-state files (#380, fixes #379) |
| v1.17.0 | Context-limit safety net (#349, fixes #346) + context budget guard (#350, fixes #347) — no more silent HTTP-400 death loops; plus nudge/exec counter alignment, tier-aware cadence reset, reasoning tokens in estimates, `/acp` error-log leak fix |
| v1.17.1 | Transform no longer scales with compression history: 13.6 s → 1.2 ms (#385, fixes #384); `qualityGate.algorithms` false warning fix (#389); fork PR builds green again (#390) |
| v1.18.0 | Adaptive compression candidates (MICRO/EPISODE targets) — opt-in via `compress.candidates`, default OFF is byte-exact v1.17.1 (#341) |
| v1.18.1 | Paper preprint v0.2 into the repo (docs only, #394) |

Note: PR #403 (Windows path-separator normalization in `protectedFilePatterns`) landed after v1.18.1 and rides the next patch release — it does not block this promotion.

## Verification

- `./scripts/ci/check-pr.sh 2026-09-17_promote-stable-v1.18.1 origin/master` — all checks passed (branch name, devlog REQ/WORKLOG present, version unchanged)
- `npm run typecheck` — clean; `npm run build` — success
- `npm test` — 1265/1267 pass; the 2 failures are local-environment artifacts (both tests hardcode `/tmp` paths, which is a read-only mount in the agent sandbox; pre-existing on master, unaffected on CI runners where `/tmp` is writable)


ranxianglei:

[bot] 🏷 Verifying PR #433 (promote-stable v1.18.1). Findings so far:

**Branch & diff** — pulled `2026-09-17_promote-stable-v1.18.1` (head `ae1ab8a7`, based on master `06efd39c`):
- Diff vs master is exactly 2 files, both additive: `devlog/2026-09-17_promote-stable-v1.18.1/REQ.md` (+28) and `WORKLOG.md` (+23). No code, config, workflow, or `package.json` changes — matches the promote-only pattern of #270/#273/#283/#309/#358.

**Mechanism** — confirmed in `.github/workflows/release.yml`:
- Commit title `promote: stable v1.18.1 — sync stable tag with latest (1.14.26 → 1.18.1)` matches **Pattern 3** (`^promote: stable v[0-9]+\.[0-9]+\.[0-9]+`) and branch name matches **Pattern 4** (`..._promote-stable-v...`), so either merge style triggers the "Promote to npm stable tag" step: `npm dist-tag add opencode-acp@1.18.1 stable`.
- The promote step is independent of the publish path — since `package.json` version is unchanged, no release/tag/npm-publish happens, consistent with the issue description.

**npm state** (live check): `stable: 1.14.26`, `latest: 1.18.1` — promotion target exists and is current.

**CI**: `pr-validation` ✅, `build` ✅, `build-artifact` ✅, `test (22)` ✅, `test (24)` ✅; `e2e` still running — I'll confirm as soon as it finishes.

Merge is a human-only operation per project rules — click merge when ready; CI will do the dist-tag update automatically. After merge, verify with `npm view opencode-acp dist-tags` → expect `stable: 1.18.1`.

初步结论：分支仅含 devlog（无代码改动、version 未动），release.yml 的 promote-stable 触发路径已核实，CI 基本全绿，等 e2e 跑完即可由人工合并。

ranxianglei:

[bot] 🏷 **CI fully green — PR #433 verified and ready to merge.**

Final check-run status on head `ae1ab8a7`:

| Check | Result |
|---|---|
| pr-validation | ✅ success |
| build | ✅ success |
| build-artifact | ✅ success |
| test (Node 22) | ✅ success |
| test (Node 24) | ✅ success |
| e2e | ✅ success |

Also re-confirmed just now: the branch still contains current `origin/master` (`06efd39c`) — no rebase needed. The 2 local test failures noted in the issue body are sandbox-only artifacts (read-only `/tmp`); both pass on CI runners as expected.

**Summary of verification:**
1. Diff = 2 devlog files only (`REQ.md` + `WORKLOG.md`); zero code/config/workflow changes; `package.json` version untouched → no changelog requirement, no npm publish triggered.
2. Merge commit will match release.yml **Pattern 3** (squash title `promote: stable v1.18.1 — …`) or **Pattern 4** (standard merge of branch `…_promote-stable-v1.18.1`) → CI runs `npm dist-tag add opencode-acp@1.18.1 stable`.
3. Live npm state pre-merge: `stable: 1.14.26`, `latest: 1.18.1`.

Next step is yours: merge the PR (human-only per project rules). After the workflow completes, verify with `npm view opencode-acp dist-tags` → expect `stable: 1.18.1`; the workflow also comments on the PR with the promotion confirmation.

中文摘要：已核对 PR 仅含 devlog、无代码与版本改动，release.yml 的 promote-stable 触发路径（Pattern 3/4）确认无误，全部 6 项 CI 检查通过且分支与 master 同步——可以合并，合并后 CI 会自动把 npm stable 标签指向 1.18.1。

